PR #7, opened 2020-12-26
More consistent naming

### LINKED ISSUE #5: EditableGridCell should have a common newValue property
```ts
type EditableGridCell = TextCell | ImageCell | BooleanCell | MarkdownCell | UriCell | NumberCell;
```

`EditableGridCell` does not have a common property across union types that represents the new value, so you have to do clumsy checking in `onCellEdited` (for example). It would be nice to be able to:

```ts
onCellEdited={([col, row], editedCell) => console.log(editedCell.newValue)}
```

### LINKED ISSUE #8: API inconsistencies
- [x] `onVisibleRowsChanged` is badly named.  It affects columns, too.  Maybe `onVisibleRegionChanged`? - yes
- [x] `onColumnMoved` takes a column index, but `onColumnResized` takes a `GridColumn`. - that's fine for now
- [x] We use an array for X/Y coordinates, a `Rectangle` object for rectangles, and `drawCustomCell` takes x/y/width/height as four separate arguments. - make `drawCustomCell` take a `Rectangle`
- [x] #5 - make the underlying data always present as `data`, with `displayData` being there sometimes
- [x] `allowInsertRow` and `onRowInserted` should probably be called "add" or "append" instead of "insert" - `showTrailingBlankRow` and `onRowAppended`
- [x] `getCellsForSelection` is passed a `GridSelection`, but only its rectangle is needed - fix it

schani:
Fixes #5 
Fixes #8


dvdsgl:
Why is the property named `displayData`? Is this the canonical cell data? I’m a bit thrown off by `display`.

schani:
Because it's what's displayed, vs what the actual underlying data is.  For example, for a number the underlying data in `editData` might be `3.1415926535` whereas `displayData` might be `"3.142"`, which is what's displayed.  But when you double-click to edit, you edit the underlying data in `editData`.

schani:
I guess the other way to do it would be to always have `editData` and for `displayData` to be only there when the two are different.  So booleans only have `editData` (we could also name it just `data`), but numbers have both, because their display can differ from their underlying value.  That required a tiny bit more than just renaming, though.

jassmith:
Please dont bump the version number until we are ready to roll a release.

PR #25, opened 2021-02-26
Add API for smooth scrolling

chkn:
Set `smoothScrollX` or `smoothScrollY` to `true` on `DataEditor` to have smooth scrolling on the respective axis(es).

Under the hood this has the following effects:

- Setting `smoothScrollX` or `smoothScrollY` to `true` on `GridScroller`, which causes
       `onVisibleRegionChanged` to fire with more granularity, passing
       `tx` and/or `ty` arguments as appropriate.

- Those are passed into `translateX` and `translateY` props on the data grid
       to specify a pixel translation on top of the cellXOffset/cellYOffset

If you don't do the above, it should work exactly as before, with the minor difference that `scrollLeft` is now clamped to be >= 0 in `ScrollRegion` update callback.

### DIFF
--- a/src/data-editor/data-editor-stories.tsx
+++ b/src/data-editor/data-editor-stories.tsx
@@ -30,7 +30,7 @@ export default {
         (fn: StoryFn<StoryFnReactReturnType>, context: StoryContext) => (
             <div style={{ overflow: "hidden" }}>
                 <BuilderThemeWrapper width={1000} height={800} context={context}>
-                    <DataEditorContainer width={500} height={500}>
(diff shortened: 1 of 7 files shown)


chkn:
@jassmith I think this puppy is ready for a look.

jassmith:
image
Bottom line still goes AWOL when scrolling down, make sure you test on a non-retina device.

jassmith (changes requested):
See earlier comment

chkn:
@jassmith  Good catch! Tested on Windows and fixed that and another issue as well.

PR #32, opened 2021-03-12
Render column by column instead of row by row, performs better

jassmith:


src/data-grid/data-grid.tsx
```diff
@@ -692,18 +700,31 @@ const DataGrid: React.FunctionComponent<Props> = p => {
...
+        let row = 0;
+        ctx.font = "13px Roboto, sans-serif";
+        for (const c of effectiveCols) {
             let y = headerHeight + translateY;
+            row = cellYOffset;
+
+            ctx.save();
+            ctx.beginPath();
+            if (c.sticky) {
+                clipX = Math.max(clipX, x + c.width);
```
invaliduser, line 715: nitpick but I don't see x change anywhere, and it's set to 0 on 704.

jassmith: god catch!

jassmith: Actually line 821 it gets set

invaliduser: mb, thought 821 was referring to (deleted) 704's x

PR #43, opened 2021-05-28
Allow a maximum column width to be specified

VictoriaLacroix:
Previously, the column size would hit a hard cap at 500 pixels.  We would like to have columns that are wider than this, so I've made it a prop.


VictoriaLacroix:
Hmm.  Installing this locally, it doesn't seem to work quite yet.  I'll investigate further.

PR #51, opened 2021-06-23
Fix crash

sethrosetter:
- Fix crash for undefined index when columns change
~~Clear grid selection when columns change~~ Not a good way to do this currently since columns are maintained by the client of the lib

### DIFF
--- a/src/data-editor/data-editor.tsx
+++ b/src/data-editor/data-editor.tsx
@@ -179,6 +179,9 @@ const DataEditor: React.FunctionComponent<DataEditorProps> = p => {
     );
 
     const mangledCols = React.useMemo(() => {
+        // Clear the grid selection to handle the case where changing columns
+        // can highlight new unexpected selections
+        setGridSelection(undefined);
         if (!rowMarkers) return columns;
         return [
             {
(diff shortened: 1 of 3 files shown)


sethrosetter:
Not sure `setGridSelection(undefined)` is the safest. I could see an edge case where the callback could mutate the columns and then there could be a possible infinite loop.

src/data-editor/data-editor.tsx
```diff
@@ -179,6 +179,9 @@ const DataEditor: React.FunctionComponent<DataEditorProps> = p => {
     );
 
     const mangledCols = React.useMemo(() => {
+        // Clear the grid selection to handle the case where changing columns
+        // can highlight new unexpected selections
+        setGridSelection(undefined);
```
ivoelbert, line 184: Woah woah woah, careful with this. The function inside `useMemo` runs at render time. We shouldn't have side effects on `useMemo`.

ivoelbert (changes requested):

src/data-editor/data-editor.tsx
sethrosetter: Removed this for now

PR #62, opened 2021-09-10
Feature/add yalo colors

andrecrts:


andrecrts:
sorry this doenst go here. cant delete it :/

PR #64, opened 2021-09-13
Improve relation bubbles style

ivoelbert:
It might look too subtle, but it works.

Here's a before and after
before
after


tristanlabbe:
👍

PR #67, opened 2021-09-29
Add readonly support for text and number cells

jassmith:


bryhasagithub (commented):
When you click/double click on the cell, you could type and it would clear visually clear the cell valu

https://user-images.githubusercontent.com/11837954/135360661-1fb14380-1829-45c6-ac66-375c0e4dcf96.mov

e

bryhasagithub (changes requested):
Just needs the comment I made to be fixed

PR #107, opened 2021-11-30
Fix text edit in firefox

ivoelbert:
Text input in firefox was broken in a few ways:
- We had a scrollbar that shouldn't be there
- The textarea didn't wrap lines
- The textarea didn't grow correctly with its value


jassmith:
Merged into 3.1.3 branch

PR #118, opened 2021-12-21
Update README.md to point to new API.md path

pzcfg:
`core` got moved inside of `packages` so these links were 404ing


jassmith:
Woops thank you.

PR #132, opened 2022-01-15
Shift data down for "Add Data to Top" story

pzcfg:
The `Add Data to Top` storybook story is broken because the data source still adds the empty row to the bottom. This shifts all of the data down a row and then adds the empty row to the top.

Is there a better way of shifting/inserting rows instead of this?


jassmith:
Not for this dummy data source no. It's just the worlds dumbest data source.

PR #134, opened 2022-01-15
Allow onRowAppended to return a number for arbitrary row insertion

pzcfg:
Currently `onRowAppended` has the return type `"top" | "bottom" | undefined`. This works well if you want to prepend or append the value, but it does not allow you to insert values at an arbitrary index.

This adds `number` to the available return types (`"top" | "bottom" | number | undefined`) and uses that value for the `scrollTo` and `setGridSelection` calls instead of being limited to `row = bottom ? rows : 0`.

This also adds a "Add Data to Middle" story to the storybook to show and test its usage.

https://user-images.githubusercontent.com/94077014/149606249-13539b46-a1b4-4b86-afcd-db0ce82de95f.mp4

### DIFF
--- a/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
@@ -659,6 +659,78 @@ export const AddDataToTop: React.VFC = () => {
     },
 };
 
+interface AddDataToMiddleProps {
+    insertIndex: number;
+}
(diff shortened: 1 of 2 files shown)


jassmith:
Sorry for the delay. I was taking a break for the holidays. Can you please fix the conflicts?

pzcfg:
done!

jassmith:
Thanks will get this in shortly

jassmith (approved):

PR #139, opened 2022-01-27
Add appendRowRef to allow appending to be triggered elsewhere

pzcfg:
I'd like to be able to trigger the append action from outside of the provided row, but there doesn't currently seem to be an exposed way to do this.

I'm not sure if this is the right way to go about it (I'm hoping there's a cleaner method) but I added an `appendRowRef` prop that gets `appendRow` assigned to it, and I am able to call it from elsewhere.

https://user-images.githubusercontent.com/94077014/151271212-a6450588-4fbb-438b-8485-9bb0c33b9065.mp4

### DIFF
--- a/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
@@ -659,6 +659,63 @@ export const AddDataToTop: React.VFC = () => {
     },
 };
 
+export const AppendRowRef: React.VFC = () => {
+    const { cols, getCellContent, setCellValueRaw, setCellValue } = useMockDataGenerator(60, false);
+
+    const [numRows, setNumRows] = React.useState(50);
+
+    const appendRowRef = React.useRef<any>(null);
+
+    const onClick = React.useCallback(() => {
+      const appendRow = appendRowRef.current;
+      appendRow(4)
+    }, [appendRowRef]);
+
+    const onRowAppended = React.useCallback(() => {
+        const newRow = numRows;
+        for (let c = 0; c < 6; c++) {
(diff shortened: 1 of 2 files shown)


ivoelbert:
Hi @pzcfg! Thanks for the PR 😄 

The preferred way to add new rows would be to add them in your data source.

If you want the scroll behavior (scroll to the added cell), we're adding support for imperatively scrolling in https://github.com/glideapps/glide-data-grid/pull/147

Is that enough for your use case?

pzcfg:
imperative scrolling would be great! is there also a way to set the selection / focus / etc. so they can start editing/typing in the cell immediately?

pzcfg:
also, would it be possible to get more control over the scrolling? I think `scrollTo` would still be valuable to have, but in some places I want to scroll further down past the cell (even the last cell, using `paddingBottom`) so that there is space for a dropdown from our custom editor

ivoelbert:
> is there also a way to set the selection / focus / etc. so they can start editing/typing in the cell immediately?

Sounds useful! We might add that to the imperative handles. We're also considering adding an imperative handle for adding a cell, so you don't have to replicate all the behavior yourself. We'd so something similar to what you provided, but adding it in `useImperativeHandle` instead of passing down a ref. We'll keep you updated on this.

> would it be possible to get more control over the scrolling? 
Sure! I'm adding that to `scrollTo`. We'll keep you updated on this as well.

ivoelbert:
Just added two optional params, `paddingX` and `paddingY` to `scrollTo`

https://github.com/glideapps/glide-data-grid/pull/150

It'll be available in the next release.

ivoelbert:
Also, I'll close this PR cause in case we implement this (which I think we will), we'll go with a different approach.

PR #145, opened 2022-02-02
Lossy Copy Data should not be a story

### LINKED ISSUE #122: "Lossy Copy Data" Story in Storybook has an error - not loading
there is an error happened in https://glideapps.github.io/glide-data-grid/?path=/story/dataeditor--lossy-copy-data.
like this: 
image

ivoelbert:
Should close #122 

Storybook picked this up as a story that just threw an error.


bryhasagithub (approved):
hahaha I love the solution

PR #153, opened 2022-02-03
Allow for unstyled custom editors

### LINKED ISSUE #123: Wrapperless cell editor
I'm building a custom cell editor using the `provideEditor` attribute, but it seems that this enforces that the editor gets put inside a `DataGridOverlayEditorStyle` and  `clip-region` div which adds its own border, border radius and styles `input` and `textarea` elements.

https://github.com/glideapps/glide-data-grid/blob/6b0a04f9d6550378890580b4db1e1168e4268c54/packages/core/src/data-grid-overlay-editor/data-grid-overlay-editor.tsx#L108-L117

Would it be possible to support a version of provideEditor that handles its own borders to allow for more flexibility?

I see that I can add a `disablePadding` property to remove the padding, but it would be great if there was another attribute that could remove some of this other styling too.

https://github.com/glideapps/glide-data-grid/blob/6b0a04f9d6550378890580b4db1e1168e4268c54/packages/core/src/data-grid-overlay-editor/data-grid-overlay-editor.tsx#L71-L82

jassmith:
fixes #123

### DIFF
--- a/packages/core/src/data-grid-overlay-editor/data-grid-overlay-editor-style.tsx
+++ b/packages/core/src/data-grid-overlay-editor/data-grid-overlay-editor-style.tsx
@@ -24,16 +24,8 @@ export const DataGridOverlayEditorStyle = styled.div<Props>`
     max-width: 400px;
     max-height: calc(100vh - ${p => p.targetRect.y + 10}px);
 
-    border-radius: 2px;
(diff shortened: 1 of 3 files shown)


ivoelbert (approved):
lgtm :shipit:

Issue #159, opened 2022-02-04
Add option to draw extra cells for paddingBottom/paddingRight


pzcfg:
Right now `paddingBottom` and `paddingRight` doesn't draw anything in the "padded" region

https://user-images.githubusercontent.com/94077014/152489535-43094683-ced4-4ce5-b676-4d7d9249172e.mp4

but in my case I'd like the empty cell lines to still be drawn in these areas.

https://user-images.githubusercontent.com/94077014/152489554-a5bee1ec-d03a-4d85-aebb-e6a67c5fcf8b.mp4

Terminology-wise, I think the first case would make more sense as `marginBottom` and `marginRight`, and `paddingBottom` and `paddingRight` would include the drawn lines. Alternatively maybe there could be some other boolean (or booleans) to control if/which of these areas are drawn in.

https://github.com/glideapps/glide-data-grid/blob/0916a62916d0e5a1ee3e5ae091d0567b1e1a34d8/packages/core/src/scrolling-data-grid/infinite-scroller.tsx#L116-L121

```
-            width: el.clientWidth - paddingRight,
-            height: el.clientHeight - paddingBottom,
+            width: el.clientWidth,
+            height: el.clientHeight,
```

bryhasagithub:
I too, like the idea!! Do you plan on or already have a PR open for the suggested addition?

pzcfg:
I can make one, but it'd be good to settle on terminology/approach first

pzcfg:
I went ahead and put up a proposal for this in #160

https://user-images.githubusercontent.com/94077014/152608583-6ab100f6-9a4a-4784-b3ba-800316b5f711.mp4

ivoelbert:
Thanks for the PR @pzcfg!

We figured that the behavior you want could be achieved without breaking changes.

Here's the PR with the relevant change. https://github.com/glideapps/glide-data-grid/pull/164

PR #160, opened 2022-02-04
Add marginBottom/marginRight and change paddingBottom/paddingRight behavior

pzcfg:
As described in #159, currently `paddingBottom` and `paddingRight` function more like a margin, where nothing is drawn in that area. This changes `paddingBottom` and `paddingRight` to still draw empty cells in those areas, and uses `marginBottom` and `marginRight` to achieve the prior behavior.

https://user-images.githubusercontent.com/94077014/152607986-ff462cef-592d-4677-b413-4fb3665cc6bd.mp4

### DIFF
--- a/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
@@ -942,7 +942,7 @@ interface AddColumnsProps {
     columnsCount: number;
 }
 
-export const AddColumns: React.VFC<AddColumnsProps> = p => {
+export const AddColumns: React.FC<AddColumnsProps> = p => {
     const { cols, getCellContent } = useMockDataGenerator(p.columnsCount);
 
     return (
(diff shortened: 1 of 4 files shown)


ivoelbert:
Hhmm, on the one hand I agree with the terminology, but on the other this means a breaking change, and a quite complex one cause we're re-purposing part of the API to do some _other thing_. Also, I don't know if it's too much of an issue cause this lives in the `experimental` prop.

We'll think about this and keep you updated. Thanks for the PR!

pzcfg:
Yeah repurposing was a concern for me too, but my thinking is that it is `experimental`, and that for most use cases, the behavior would essentially be identical: if they're completely drawing over the padded area, it will look and scroll the same as it did before, you'll just happen to be drawing a few extra lines underneath. If they aren't completely covering this area (e.g. opacity, blur, inset overlay) then this is likely the preferred/desired behavior. Either way it is technically a breaking change though and should probably be called out in the release notes.

ivoelbert:
Hey there! I'm about to add support for `overscrollY`, which in combination with `overscrollX` should let you do this exact same thing without breaking changes.

https://github.com/glideapps/glide-data-grid/pull/164

ivoelbert:
Hey there! I'm about to add support for `overscrollY`, which in combination with `overscrollX` should let you do this exact same thing without breaking changes.

https://github.com/glideapps/glide-data-grid/pull/164

PR #167, opened 2022-02-10
Update padding when paddingBottom and paddingRight values change

pzcfg:
Currently the `paddingBottom` and `paddingRight` values don't update until you interact with the grid.

https://user-images.githubusercontent.com/94077014/153515433-a32f2355-91d7-4cb8-8d77-e960c129d831.mp4

This fixes that by calling `onScroll` whenever `paddingBottom` or `paddingRight` change

https://user-images.githubusercontent.com/94077014/153515458-9d13fa6a-9ad6-43a6-bf44-09f99db7cd66.mp4

I did notice that when this value is changing rapidly the width/height of the cells seems to briefly change. Not sure what is causing that

https://user-images.githubusercontent.com/94077014/153515561-204abf3b-cb2c-435f-b5ea-4ddb57c7d812.mp4

### DIFF
--- a/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/data-editor-beautiful.stories.tsx
@@ -571,6 +571,63 @@ export const Overscroll: React.VFC<OverscrollProps> = p => {
     },
 };
 
+
+interface PaddingProps {
+    paddingRight: number;
+    paddingBottom: number;
+}
+
+export const Padding: React.VFC<PaddingProps> = p => {
+    const { paddingRight, paddingBottom } = p;
+    const { cols, getCellContent } = useMockDataGenerator(20);
+
+    return (
+        <BeautifulWrapper
+            title="Padding"
+            description={
+                <>
+                    <Description>
+                        You can add padding at the ends of the grid by setting the{" "}
+                        <PropName>paddingRight</PropName> and <PropName>paddingBottom</PropName> props
+                    </Description>
+                </>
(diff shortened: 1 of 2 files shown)


ivoelbert:
Great catch! I'll look into this a bit more. Usually having effects with "fake" dependencies (paddingX is not used in that effect, we just want the re render) is asking for trouble. We're very guilty of that, as well.

Thanks for reporting, and double thanks for sending a PR.

jassmith:
I suspect the jitter is 100% due to onScroll not forcing things to draw unblitted. The blitting is thus stretched.

jassmith:
Okay I spent more time digging into this. The fluttering is due to a desync between the DOM element resizing and the redraw happening. I am honestly not entirely sure how to fix this.

jassmith:
Okay I figured it out. This will require a lot of testing because it fundamentally changes how we size the canvas. https://github.com/glideapps/glide-data-grid/tree/fix-padding-resize Give it a try see if you like the solution and consider including it in your branch.

jassmith (changes requested):
This PR needs to incorporate some version of the modification I made in the demo branch + some relatively extensive testing of different embedding scenarios to ensure behavior remains the same.

jassmith:
Hey amigo, I'd really like to bring this in to 3.4.0 but if I just merge my branch you wont get the credit you deserve. I'd love if you took a moment to check out what can be done here.

jassmith:
Fixed in 3.4.0

Issue #158, opened 2022-02-04
Use font styling from theme/themeOverride for input fields


pzcfg:
Right now it seems that the input font size is hardcoded to 13px

https://github.com/glideapps/glide-data-grid/blob/6b0a04f9d6550378890580b4db1e1168e4268c54/packages/core/src/data-grid-overlay-editor/private/number-overlay-editor-style.tsx#L3-L14

https://github.com/glideapps/glide-data-grid/blob/6b0a04f9d6550378890580b4db1e1168e4268c54/packages/core/src/growing-entry/growing-entry-style.tsx#L1-L9

<img width="155" alt="Screen Shot 2022-02-03 at 7 59 34 PM" src="https://user-images.githubusercontent.com/94077014/152469739-20ec67cc-70af-4d9a-a317-696f8ab1b59b.png">

It would be great if this could match the theme/themeOverride of the cells so it's as WYSIWYG as possible, at least optionally, or customized separately

jassmith:
great idea let me see what can be done

jassmith:
Merged

PR #174, opened 2022-02-15
font size style theme

krisolchova:
Issue fix "Use font styling from theme/themeOverride for input fields" #158

### DIFF
--- a/packages/core/src/common/styles.ts
+++ b/packages/core/src/common/styles.ts
@@ -39,6 +39,7 @@ export interface Theme {
     headerFontStyle: string;
     baseFontStyle: string;
     fontFamily: string;
+    fontSize: string;
 }
 
 const dataEditorBaseTheme: Theme = {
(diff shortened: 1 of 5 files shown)


packages/core/src/common/styles.ts
```diff
@@ -39,6 +39,7 @@ export interface Theme {
     headerFontStyle: string;
     baseFontStyle: string;
     fontFamily: string;
+    fontSize: string;
```
jassmith, line 42: `editorFontSize` would be a better name here as it may be unclear what the intent of this theme property is otherwise.

jassmith (changes requested):

packages/core/src/common/styles.ts
krisolchova: yep makes sense.

bryhasagithub (approved):
Hooray! Thanks for making the change

PR #201, opened 2022-03-13
Implement basic dropdown cell and article cell

jassmith:


package.json
```diff
@@ -34,8 +34,8 @@
         "@types/faker": "^5.5.8",
         "@types/lodash": "^4.14.177",
         "@types/marked": "^4.0.2",
-        "@types/react": "^16.14.21",
-        "@types/react-dom": "^16.9.5",
+        "@types/react": "^16.14.21 || 17.x",
+        "@types/react-dom": "^16.9.5 || 17.x",
```
ivoelbert, line 38: woooo fancy

packages/cells/src/cells/dropdown-cell.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+interface CustomMenuProps extends MenuProps<any> {}
+
+export class CustomMenu extends React.Component<CustomMenuProps> {
+    render() {
+        const { Menu } = components;
+        const { children, ...props } = this.props;
+        return <Menu {...props}>{children}</Menu>;
+    }
+}
```
ivoelbert, line 15: Why is this a class, lol

packages/cells/src/cells/dropdown-cell.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export type DropdownCell = CustomCell<DropdownCellProps>;
+
+const Wrap = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
+`;
+
+const editor: ProvideEditorCallback<DropdownCell> = () => {
```
ivoelbert, line 32: I'm pretty sure you can do
```ts
const InnerEditor: ReturnType<ProvideEditorCallback<DropdownCell>> = p => {
    // ...
}

InnerEditor.disablePadding = true

const editor: ProvideEditorCallback<DropdownCell> = () => InnerEditor;
```

This way `editor` will now return a stable reference all the time, and you won't have to ignore `rules-of-hooks`

ivoelbert (commented):
I think this alternative is reasonable

packages/cells/src/cells/dropdown-cell.tsx
jassmith: Followed an example blindly.

jassmith: Fixed

packages/cells/src/cells/dropdown-cell.tsx
jassmith: You are right, fixed

PR #202, opened 2022-03-13
Prevent spurious event emission

### LINKED ISSUE #196: Change callbacks trigger even if there was no change
The change callbacks - `onSelectedColumnsChange`, `onSelectedRowsChange`, `onGridSelectionChange` - get called also when there is no change in the respective selection. For example, `onSelectedRowsChange` gets called whenever a cell is selected,  a column is selected or escape is pressed even so no row was selected before. There is an easy workaround by checking against the current selection state (`selectedRows`, `selectedColumns`, `gridSelection`). However, I don't know if that is indeed the desired behavior? Maybe this check with the actual selection state could be implemented directly into glide-data-grid?

Btw. `onGridSelectionChange` is currently not documented in the API docs. Is it safe to use this callback?

jassmith:
Fixes #196

### DIFF
--- a/packages/core/src/data-grid/data-grid-types.ts
+++ b/packages/core/src/data-grid/data-grid-types.ts
@@ -500,7 +500,7 @@ export class CompactSelection {
             const left = other.items[i];
             const right = this.items[i];
 
-            if (left[0] !== right[0] || left[1] || right[1]) return false;
+            if (left[0] !== right[0] || left[1] !== right[1]) return false;
         }
 
         return true;
--- a/packages/core/src/data-grid/data-grid-types.ts
+++ b/packages/core/src/data-grid/data-grid-types.ts
@@ -491,6 +491,21 @@ export class CompactSelection {
         return false;
     };
 
+    equals = (other: CompactSelection): boolean => {
+        if (other === this) return true;
+
+        if (other.items.length !== this.items.length) return false;
+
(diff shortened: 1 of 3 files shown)


packages/core/src/data-grid/data-grid-types.ts
```diff
@@ -491,6 +491,21 @@ export class CompactSelection {
...
+    equals = (other: CompactSelection): boolean => {
+        if (other === this) return true;
+
+        if (other.items.length !== this.items.length) return false;
+
+        for (let i = 0; i < this.items.length; i++) {
+            const left = other.items[i];
+            const right = this.items[i];
+
+            if (left[0] !== right[0] || left[1] || right[1]) return false;
```
ivoelbert, line 503: Pretty sure you meant
```ts
if (left[0] !== right[0] || left[1] !== right[1]) return false;
```

ivoelbert (changes requested):

packages/core/src/data-grid/data-grid-types.ts
jassmith: Guess I need to figure out why this didn't cause test failures either... it definitely should have

ivoelbert: eslint should prevent this

PR #203, opened 2022-03-13
Add onSelectionCleared

### LINKED ISSUE #197: Add callback to detect the clearing of all selections
For my use case, I need to react to the event when all selections are cleared (when escape is pressed). This turned out to be a bit complicated to implement with glide-data-grid, since this event currently only triggers all individual selection change callbacks. I struggled a bit with detecting that indeed all selections were cleared on the basis of the individual callbacks. My hacky workaround was to also react to the escape key.

A solution for this might be to add an `onSelectionsCleared` callback that is called when escape is pressed. In addition, it might be useful to add a way to also trigger the selection clearing via the API - e.g. `clearSelections` - similar to how `showSearch` works. This could be useful to implement a clearing action in a table toolbar or on mobile where we don't have the escape hotkey.

jassmith:
fixes #197

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -937,6 +937,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 focus();
                 setSelectedColumns(CompactSelection.empty(), "selection");
                 setSelectedRows(CompactSelection.empty());
+                onSelectionCleared?.();
                 lastSelectedRowRef.current = undefined;
                 lastSelectedColRef.current = undefined;
             } else if (args.kind === "group-header") {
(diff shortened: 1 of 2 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1527,6 +1530,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     setGridSelection(undefined, false);
                     setSelectedRows(CompactSelection.empty());
                     setSelectedColumns(CompactSelection.empty(), "selection");
+                    onSelectionCleared?.();
```
ivoelbert, line 1533: Shouldn't we call this when we click out-of-bounds as well?
Around here:
https://github.com/glideapps/glide-data-grid/blob/6b857686c061ca8d9e3fffca4135356625b4ac63/packages/core/src/data-editor/data-editor.tsx#L939

ivoelbert (commented):
I think you missed a case there

packages/core/src/data-editor/data-editor.tsx
jassmith: Good catch

PR #243, opened 2022-04-02
rowMarkerOffset for onCellActivated mouse clicks

reczy:
Keyboard activation already accounts for rowMarkerOffset; this fixes for mouse clicks


jassmith:
Thanks

PR #257, opened 2022-04-12
add props for onColumnResizedStart and onColumnResizedEnd

BrianHung:
https://github.com/glideapps/glide-data-grid/discussions/254

Modeled off
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ondrag
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ondragstart
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ondragend

### DIFF
--- a/packages/core/src/data-grid-dnd/data-grid-dnd.tsx
+++ b/packages/core/src/data-grid-dnd/data-grid-dnd.tsx
@@ -10,7 +10,8 @@ export interface DataGridDndProps extends Props {
     readonly onColumnMoved?: (startIndex: number, endIndex: number) => void;
     readonly onColumnResized?: (column: GridColumn, newSize: number) => void;
     readonly onColumnResizedStart?: (column: GridColumn, newSize: number) => void;
-    readonly onColumnResizedEnd?: (column: GridColumn, newSize: number) => void;    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
+    readonly onColumnResizedEnd?: (column: GridColumn, newSize: number) => void;
+    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
     readonly maxColumnWidth: number;
     readonly minColumnWidth: number;
     readonly lockColumns: number;
--- a/packages/core/src/data-grid-dnd/data-grid-dnd.tsx
+++ b/packages/core/src/data-grid-dnd/data-grid-dnd.tsx
@@ -9,7 +9,8 @@ export interface DataGridDndProps extends Props {
     readonly onRowMoved?: (startIndex: number, endIndex: number) => void;
     readonly onColumnMoved?: (startIndex: number, endIndex: number) => void;
     readonly onColumnResized?: (column: GridColumn, newSize: number) => void;
-    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
+    readonly onColumnResizedStart?: (column: GridColumn, newSize: number) => void;
+    readonly onColumnResizedEnd?: (column: GridColumn, newSize: number) => void;    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
     readonly maxColumnWidth: number;
     readonly minColumnWidth: number;
     readonly lockColumns: number;
@@ -32,6 +33,8 @@ const DataGridDnd: React.FunctionComponent<DataGridDndProps> = p => {
     const {
         onColumnMoved,
         onColumnResized,
+        onColumnResizedStart,
+        onColumnResizedEnd,
         gridRef,
         maxColumnWidth,
         minColumnWidth,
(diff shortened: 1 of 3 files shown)


packages/core/src/data-grid-dnd/data-grid-dnd.tsx
```diff
@@ -9,7 +9,8 @@ export interface DataGridDndProps extends Props {
     readonly onRowMoved?: (startIndex: number, endIndex: number) => void;
     readonly onColumnMoved?: (startIndex: number, endIndex: number) => void;
     readonly onColumnResized?: (column: GridColumn, newSize: number) => void;
-    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
+    readonly onColumnResizedStart?: (column: GridColumn, newSize: number) => void;
+    readonly onColumnResizedEnd?: (column: GridColumn, newSize: number) => void;    readonly gridRef?: React.MutableRefObject<DataGridRef | null>;
```
jassmith, line 13: missing newline character

packages/core/src/data-grid-dnd/data-grid-dnd.tsx
```diff
@@ -32,6 +33,8 @@ const DataGridDnd: React.FunctionComponent<DataGridDndProps> = p => {
     const {
         onColumnMoved,
         onColumnResized,
+        onColumnResizedStart,
```
jassmith, line 36: These need to be mangled in data-editor.tsx so they account for the rowMarkerOffset.

jassmith (changes requested):

packages/core/src/data-grid-dnd/data-grid-dnd.tsx
BrianHung: Could you elaborate what you mean by that? 

The normal `onColumnResized` doesn't seem to be used in `data-editor.tsx`.
https://github.com/glideapps/glide-data-grid/blob/main/packages/core/src/data-editor/data-editor.tsx

For `onColumnResizedStart`, I just use the `args.bounds.width`. For `onColumnResizedEnd`, ` lastResizeWidthRef.current`.

jassmith: Sorry my brain was on autopilot here. This takes the actual column not the index. This presents a different issue of sorts but one that is pre-existing. The auto column sizer will now make this feed back a different object that what was fed in by the user. Thought is needed here, but not a blocker for this PR. Can you please retarget to 4.0.3?

BrianHung:
@jassmith Jw, what is the expected behavior these two lines are implementing? I understand theyre for multi-col resizing, but why does `onMouseMove` check for `selectedColumns?.first() === resizeCol` while `onMouseUp` checks for `electedColumns?.hasIndex(resizeCol) === true` ?

https://github.com/glideapps/glide-data-grid/blob/0e62b1cdf86d30e8e7aedc733d2f338f21b3b62b/packages/core/src/data-grid-dnd/data-grid-dnd.tsx#L115-L120

https://github.com/glideapps/glide-data-grid/blob/0e62b1cdf86d30e8e7aedc733d2f338f21b3b62b/packages/core/src/data-grid-dnd/data-grid-dnd.tsx#L185-L191

BrianHung:
@jassmith Heads up since this got merged into branch 4.0.3, after revisiting it a bit more, I noticed the PR didn't correctly handle 

1. `onColumnResizedStart` for `selectedColumns`, and 
2. `onColumnResizedEnd` doesn't get called if `selectedColumns` is empty. 

(Sorry, shouldve marked it as a draft until I fully tested everything! 😅)

---

https://github.com/glideapps/glide-data-grid/blob/b8838978f5b0efb2684ac1d888952025bccc1aa9/packages/core/src/data-grid-dnd/data-grid-dnd.tsx#L81-L86

should instead be (?) to handle the `selectedColumns` case:

```ts
 if (args.isEdge) { 
    shouldFireEvent = false;
    setResizeColStartX(args.bounds.x);
    setResizeCol(col);
    onColumnResizedStart?.(columns[col], args.bounds.width)
    for (const c of selectedColumns) {
       if (c === col) continue;
       onColumnResizedStart?.(columns[c], args.bounds.width)
}
```

---

https://github.com/glideapps/glide-data-grid/blob/b8838978f5b0efb2684ac1d888952025bccc1aa9/packages/core/src/data-grid-dnd/data-grid-dnd.tsx#L117-L129

should be changed to call `onColumnResizedEnd` when `selectedColumns` is empty:

```ts
if (resizeCol !== undefined && resizeColStartX !== undefined) {
    const column = columns[resizeCol];
    const newWidth = clamp(lastResizeWidthRef.current, minColumnWidth, maxColumnWidth);
    onColumnResizedEnd?.(column, newWidth);

    if (selectedColumns?.hasIndex(resizeCol) === true) {
        for (const c of selectedColumns) {
            if (c === resizeCol) continue;
            onColumnResizedEnd?.(columns[c], newWidth);
        }
    }
}
```

---

Last thing, in the second `if` condition, `resizeCol !== undefined` can be removed since the first `if` condition already checks for that.

https://github.com/glideapps/glide-data-grid/blob/b8838978f5b0efb2684ac1d888952025bccc1aa9/packages/core/src/data-grid-dnd/data-grid-dnd.tsx#L189-L201

jassmith:
thanks I already made some fixes I will incorporate these changes. Im also writing some tests for it

jassmith:
it didn't go into main yet, just the feature branch where I am much free-er with bringing in work from others.

PR #310, opened 2022-04-29
Support indeterminate and empty states for boolean cell

### LINKED ISSUE #309: Support for null/undefined value in boolean cell
The boolean cell currently is only able to show a checked or unchecked checkbox. However, in case the cell value is actually null or undefined, it would be great if it would hide the checkbox. This is useful for tables that contain missing data.

ivoelbert:
Should close #309 

New styles look like:

image

Bonus:
I added a `focus` imperative handle for the DataEditor that just forwards to DataGrid

### DIFF
--- a/packages/core/src/data-editor/data-editors-fns.ts
+++ b/packages/core/src/data-editor/data-editors-fns.ts
@@ -1,6 +1,6 @@
 import { assertNever } from "../common/support";
 import { DataGridSearchProps } from "../data-grid-search/data-grid-search";
-import { GridCell, GridCellKind, GridSelection, Rectangle } from "../data-grid/data-grid-types";
+import { BooleanEmpty, BooleanIndeterminate, GridCell, GridCellKind, GridSelection, Rectangle } from "../data-grid/data-grid-types";
 
 export function expandSelection(
     newVal: GridSelection,
@@ -173,7 +173,7 @@ export function copyToClipboard(cells: readonly (readonly GridCell[])[], columnI
         return str;
     }
 
-    const formatBoolean = (val: boolean | null | undefined): string => {
+    const formatBoolean = (val: boolean | BooleanEmpty | BooleanIndeterminate): string => {
         switch (val) {
             case true:
                 return "TRUE";
@@ -181,10 +181,10 @@ export function copyToClipboard(cells: readonly (readonly GridCell[])[], columnI
             case false:
                 return "FALSE";
 
-            case undefined:
+            case BooleanIndeterminate:
                 return "INDETERMINATE";
 
-            case null:
+            case BooleanEmpty:
                 return "";
 
             default:
--- a/packages/core/src/data-editor/data-editors-fns.ts
+++ b/packages/core/src/data-editor/data-editors-fns.ts
@@ -185,7 +185,7 @@ export function copyToClipboard(cells: readonly (readonly GridCell[])[], columnI
                 return "INDETERMINATE";
 
             case null:
-                return "EMPTY";
+                return "";
 
             default:
                 assertNever(val);
--- a/packages/core/src/data-editor/data-editors-fns.ts
+++ b/packages/core/src/data-editor/data-editors-fns.ts
@@ -173,6 +173,25 @@ export function copyToClipboard(cells: readonly (readonly GridCell[])[], columnI
         return str;
     }
 
+    const formatBoolean = (val: boolean | null | undefined): string => {
+        switch (val) {
+            case true:
+                return "TRUE";
+
+            case false:
+                return "FALSE";
+
+            case undefined:
+                return "INDETERMINATE";
+
+            case null:
+                return "EMPTY";
+
+            default:
+                assertNever(val);
+        }
+    };
+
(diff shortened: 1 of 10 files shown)


packages/core/src/data-editor/data-editors-fns.ts
```diff
@@ -173,6 +173,25 @@ export function copyToClipboard(cells: readonly (readonly GridCell[])[], columnI
...
+                return "TRUE";
+
+            case false:
+                return "FALSE";
+
+            case undefined:
+                return "INDETERMINATE";
+
+            case null:
+                return "EMPTY";
```
jassmith, line 188: I would argue null should format as FALSE or an empty string.

packages/core/src/data-grid/cells/boolean-cell.tsx
```diff
@@ -1,7 +1,31 @@
...
 import { GridCellKind, BooleanCell } from "../data-grid-types";
 import { InternalCellRenderer } from "./cell-types";
 
+/**
+ * Checkbox behavior:
+ *
+ * true + click -> unchecked
+ * false + click -> checked
+ * indeterminate + click -> checked
+ * empty + click -> empty
```
jassmith, line 12: This behavior is achieved by simply dumping out a loadingCell instead. Why not have this act like a checkbox which is "empty" but when hovered it shows the empty state and when clicked becomes checked. This provides new behavior instead of reproducing already possible behavior.

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -377,30 +378,56 @@ function drawCheckbox(
...
-        ctx.beginPath();
-        roundedRect(ctx, centerX - 9, centerY - 9, 18, 18, 4);
+    switch (checked) {
+        case true: {
+            ctx.beginPath();
+            roundedRect(ctx, centerX - 9, centerY - 9, 18, 18, 4);
 
-        ctx.fillStyle = highlighted ? theme.accentColor : theme.textLight;
-        ctx.fill();
+            ctx.fillStyle = highlighted ? theme.accentColor : theme.textMedium;
```
jassmith, line 386: Why medium?

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -377,30 +378,56 @@ function drawCheckbox(
...
+            ctx.beginPath();
+            roundedRect(ctx, centerX - 8.5, centerY - 8.5, 17, 17, 4);
+
+            ctx.lineWidth = 1;
+            ctx.strokeStyle = hovered ? theme.textDark : theme.textMedium;
+            ctx.stroke();
+            break;
+        }
+
+        case undefined: {
```
jassmith, line 412: again I think here null should draw the same as false IF you are hovered.

jassmith (changes requested):

packages/core/src/data-grid/data-grid-lib.ts
ivoelbert: 2 folded:
- I needed a pale color for the indeterminate state that was easy to distinguish from the checked state.
- This bumps contrast and still looks nice

packages/core/src/data-editor/data-editors-fns.ts
ivoelbert: Hhmm, yeah perhaps `""`.

packages/core/src/data-grid/cells/boolean-cell.tsx
ivoelbert: Huh, I didn't think about the loading cell. I agree, then.

I'll share this option with the issue's creator.

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -510,11 +537,16 @@ function roundedRect(
...
+export function drawBoolean(args: BaseDrawArgs, data: boolean | null | undefined, canEdit: boolean) {
+    if (data === null) {
+        return;
+    }
+
     const { ctx, hoverAmount, theme, x, y, w, h, highlighted, hoverX, hoverY } = args;
+
     const hoverEffect = 0.35;
 
-    ctx.globalAlpha = canEdit ? 1 - hoverEffect + hoverEffect * hoverAmount : 0.4;
```
ivoelbert, line 517: Wrong change. Reverting this one.

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -510,11 +538,20 @@ function roundedRect(
...
+    let alpha = canEdit ? 1 - hoverEffect + hoverEffect * hoverAmount : 0.4;
+    if (data === null) {
+        alpha *= hoverAmount;
+    }
+    ctx.globalAlpha = alpha;
```
jassmith, line 554: if alpha === 0 return early. Let's not waste the draw calls drawing nothing. Unfortunately the canvas API will faithfully perform these useless operations.

ivoelbert: Ooohhh good call. Missed that one.

PR #311, opened 2022-04-29
Remove showUnchecked property from boolean cell

### LINKED ISSUE #308: Unused `showUnchecked` property of boolean cell
The `showUnchecked` property of boolean cell seems to not have any actual usage. It is only used in some tests/stories:

https://github.com/glideapps/glide-data-grid/blob/fabca8608293d9649cc7b45d503078a9192c6d7f/packages/core/src/data-grid/data-grid-types.ts#L391

ivoelbert:
Should close #308 

Not sure why that property is there, but it's not used anywhere.

Also, is this a breaking change? TS users will get errors (albeit trivially solved) but there are no behavior changes 🤔


jassmith:
Already fixed with deprecation in 4.0.4. The deprecation has no runtime cost anyway as its typescript only and prevents breaking working code.

PR #324, opened 2022-05-09
4.1.0

jassmith:
# 4.1.0 Release Notes

## 🚨 Deprecations

- `BooleanCell.showUnchecked` now deprecated and replaced with passing null as the BooleanCell value.
- `BooleanCell.allowEdit` now deprecated and replace with `BooleanCell.readonly`.

Both deprecated options continue to work as intended, but implementors should begin migrating to the new options.

### 🚣‍♂️ Improved support for styling of trailing row

`trailingRowOptions` of a `GridColumn` may now include a `themeOverride` and a `disabled` boolean which do what they say.

### ✅⁉️ Boolean cell supports indeterminate and empty states

image

`BooleanCell` can now take `null` or `undefined` as its `data` to show empty or indeterminate states respectively. To make this easier to remember, named exports are included for `BooleanIndeterminate` and `BooleanEmpty`.

### 👉👈 Support for overriding text alignment in cells

All `GridCell` types now support a `contentAlign` parameter which can be set to `"right" | "left" | "center"` and will cause the text in the cell to be aligned as requested. Thanks to @LukasMasuch 

### 🪥 Custom cells can now handle Paste with ease

Custom cell renders can now optionally provide an `onPaste` callback which receives the `CustomCells` data property and a paste string. Implementors should return back a new `data` value which will then be used to create a new cell to pass to `onCellEdited`. This allows `onPaste` for custom cells to work the same as it does for all built in cells.

All cells shipped in `glide-data-grid-cells` have been updated to support the new mechanism.

### 🐳 Mass paste handling made easier

`onCellsEdited` will be invoked if multiple cells are to be pasted into. This receives a list of all cells which are about to receive an `onCellEdited` callback due to the paste or other mass edit event. This is intended to allow more efficient handling to implementors and minimize render cycles.

### 🕺 Basic animation support for custom cells

During a draw event for a custom cell, the cell may now call `args.requestAnimationFrame()` out of the passed args to the draw event. This will trigger the cell to be drawn again in the near term future for doing animations. Draw rate automatically throttles to around 30FPS after a fixed time with no breaks in animation.

## 🐜💪 Minor Improvements

- `CompactSelection.remove` now supports removing a slice instead of just an index. Thanks to @BrianHung 
- Custom editors no longer automatically commit the tempValue when finished editing is called, allowing for editing to be properly cancelled by the custom editor.
- Text entry into a cell now more eagerly being edit mode. This makes things like entering a negative number much easier and improves foreign language support.
- Top and right edges of the grid will no longer draw a double border in some cases, thank you @LukasMasuch 
- Visually center text with better algorithm which accounts for oddities from font to font
- Boolean cells now supported in paste.

## 🪲 Bug Fixes

- Slight improvements to the newline handling for overlay editors.
- Custom cells are now properly included in search results via the `copyData` property.
- Values copied out of the grid now copy raw values into HTML and display values into text. This improves copy/paste interop inside of Glide Data Grid.
- Ideal height calculation for Grid now properly accounts for trailing row.
- Focus will no longer be lost when interacting with cells outside of the visual viewport, thank you @BrianHung 
- Fix `isDraggable`


packages/core/src/data-grid/cells/text-cell.tsx
```diff
@@ -12,7 +12,7 @@ export const textCellRenderer: InternalCellRenderer<TextCell> = {
     needsHoverPosition: false,
     renderPrep: prepTextCell,
     useLabel: true,
-    render: a => drawTextCell(a, a.cell.displayData),
+    render: a => drawTextCell(a, a.cell.displayData, a.cell.contentAlign),
```
LukasMasuch, line 15: I believe in this case only text cell would support content alignment. It would be useful to have this also activated on all the other cells using `drawTextCell` (especially number cell). The cells that currently use `drawTextCell`:
- number-cell
- markdown-cell
- row-id-cell
- text-cell
- uri-cell

PR #331, opened 2022-05-18
Add support for text alignment in theme

### LINKED ISSUE #315: Allow configuration of text alignment for number/text cells
Numbers are usually right aligned in spreadsheet software (e.g. Excel and Google Sheets) since it makes it easier to compare numbers:

image

 I would love to have an option in glide-data-grid as well to right-align the rendered text in a cell. I build my own custom cell to do this using the `textAlign` canvas configuration, but I love to have this integrated directly into the core of glide-data-grid. I think there are two good options for this:

- Option 1 (preferred): Provide the text alignment as part of the theme.  This would allow the most flexibility since the theme can be already overwritten for any column, row, or individual cell.

- Option 2: Provide an alignment property for `NumberCell` (or other text-based cells) which is applied to that cell.

LukasMasuch:
This PR adds support for changing the text alignment for number/text cells via the theme.

Closes #315

### DIFF
--- a/packages/core/src/common/styles.ts
+++ b/packages/core/src/common/styles.ts
@@ -9,7 +9,6 @@ export interface Theme {
     textMedium: string;
     textLight: string;
     textBubble: string;
-    textAlign?: "left" | "right" | "center" | "start" | "end";
 
     bgIconHeader: string;
     fgIconHeader: string;
--- a/packages/core/src/common/styles.ts
+++ b/packages/core/src/common/styles.ts
@@ -9,6 +9,7 @@ export interface Theme {
     textMedium: string;
     textLight: string;
     textBubble: string;
+    textAlign?: "left" | "right" | "center" | "start" | "end";
 
     bgIconHeader: string;
     fgIconHeader: string;
--- a/packages/core/src/data-grid/data-grid-lib.ts
+++ b/packages/core/src/data-grid/data-grid-lib.ts
@@ -7,10 +7,13 @@ import {
     SizedGridColumn,
     Rectangle,
     BaseGridCell,
+    BooleanEmpty,
+    BooleanIndeterminate,
 } from "./data-grid-types";
 import { degreesToRadians, direction } from "../common/utils";
 import React from "react";
 import { BaseDrawArgs, PrepResult } from "./cells/cell-types";
+import { assertNever } from "../common/support";
 
 export interface MappedGridColumn extends SizedGridColumn {
     sourceIndex: number;
@@ -385,7 +388,7 @@ export function drawNewRowCell(args: BaseDrawArgs, data: string, icon?: string) 
 function drawCheckbox(
     ctx: CanvasRenderingContext2D,
     theme: Theme,
-    checked: boolean,
+    checked: boolean | BooleanEmpty | BooleanIndeterminate,
     x: number,
     y: number,
     width: number,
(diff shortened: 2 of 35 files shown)


jassmith:
This isn't exactly right. Dir and text alignment produce slightly different results when it comes to overflow

```
|  ltr-start                |
|        ltr-center         |
|                   ltr-end |

// with overflow
|  ltr-start but the text ov|
|  ltr-center but the text o|
|  ltr-end but the text over|

// rtl with overflow (but using backwards english cuz)
|vo txet eht tub trats-ltr  |
```

LukasMasuch:
Here is what I get with the implementation from this PR: 

LTR + Start/Left:
image

RTL + Start/Left:
image

LTR + Center:

image

RTL + Center: 
image

LTR + End/Right:
image

RTL + End/Right:
image

-> Which seems to be equal to the current way on how RTL text is rendered:
image

If no `textAlign` is configured in the theme:

LTR text (equals LTR + Start/Left):
image

RTL text (equals RTL + End/Right):
image

Overall I think this is how I would expect text-align to work.

LukasMasuch:
You are probably referring to how text-align and overflow are handled in HTML/CSS. E.g. text-align=center will not really center if there is an overflow. However, I think this is not really what I would expect within a table cell. For example, Google Sheets is also handling alignment the same way as this PR: 

image

jassmith:
I was about to say that seems wrong but I guess its the wild fucking west out there anyway

image

jassmith:
Same thing in google sheets

image

So yeah, I will be accepting this PR. Crying a bit at the state of the world but accepting it.

packages/core/src/common/styles.ts
```diff
@@ -9,6 +9,7 @@ export interface Theme {
     textMedium: string;
     textLight: string;
     textBubble: string;
+    textAlign?: "left" | "right" | "center" | "start" | "end";
```
jassmith, line 12: I dont think this should be on the theme but a part of BaseCell. Reason for that is theme overrides are rather expensive. Props on base cell are cheap.

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -307,11 +307,17 @@ export function drawTextCell(args: BaseDrawArgs, data: string) {
...
+            // Set text align to end as default for RTL text
+            ctx.textAlign = "end";
+        }
 
-        if (dir === "rtl") {
-            const textWidth = measureTextCached(data, ctx, `${theme.baseFontStyle} ${theme.fontFamily}`).width;
-            ctx.fillText(data, x + w - theme.cellHorizontalPadding - textWidth + 0.5, y + h / 2);
+        if (ctx.textAlign === "right" || ctx.textAlign === "end") {
+            ctx.fillText(data, x + w - (theme.cellHorizontalPadding + 0.5), y + h / 2);
+        } else if (ctx.textAlign === "center") {
```
jassmith, line 319: Reading out of the state machine like this is super expensive and round trips in and out of JS. Cache this value.

jassmith (changes requested):

jassmith:
A better approach from a drawing performance standpoint will be to assume the text-align is set to start, if you encounter a situation where you need it set to something which is not start, set it to that value, then before you exit the draw call restore it back to start. Do not use save()/restore() unless needed in the cell render methods as the overhead there is quite high as well.

packages/core/src/common/styles.ts
LukasMasuch: Currently, the `drawTextCell` does not have access to the cell props, or? Should I pass the info to the `drawTextCell`, `prepTextCell` or handle it similar to how `cell.style` is handled here: https://github.com/glideapps/glide-data-grid/blob/f3f3be6d0aeb4e06fce47d51c44f724ff525f348/packages/core/src/data-grid/data-grid-render.tsx#L1237

packages/core/src/data-grid/data-grid-lib.ts
LukasMasuch: Good to know, I will change this.

packages/core/src/common/styles.ts
jassmith: pass it to the `drawTextCell`, do not pass it to `prepTextCell`

LukasMasuch: Two questions: 1) Is the theme merge also expensive if it is applied on the column instead of every individual cell? 2) If this is a cell prop, we could also make it slightly more generic and call it `contentAlign` so other cell types can also theoretically use it for aligning the content. Would you prefer `textAlign` or `contentAlign`?

packages/core/src/data-grid/data-grid-lib.ts
LukasMasuch: The expensive call here is checking the property state in `ctx`, or? So, checking the value in `theme` or cell props would not be expensive?

jassmith: Correct. Reading out of ctx is expensive if done frequently (this is frequent). Setting anything in ctx can often be orders of magnitude more expensive than you might expect. Changing the text alignment as very very not cheap and thus should be done as minimally as possible.

packages/core/src/data-grid/data-grid-lib.ts
```diff
@@ -294,7 +302,7 @@ export function prepTextCell(
     return result;
 }
 
-export function drawTextCell(args: BaseDrawArgs, data: string) {
+export function drawTextCell(args: BaseDrawArgs, data: string, cell: BaseGridCell) {
```
jassmith, line 305: I'd rather this pass just the data and contentAlign. Passing the entire cell passes more data than we potentially want for these functions to have. Minor issue but this is just a general preference. Then again... some functions in this project have just stupid numbers of args... so what do I know.

jassmith:
Still has some minor things needing fixing but I can do that post merge.

faheemhameed:
Hi there,

The textAlign property doesn't seem to work

I am using the version: @glideapps/glide-data-grid@4.1.0-beta1

Can you please check this or I am overlooking this?

Many thanks.

jassmith:
What cell are you using it in? Can you show me a demo of it not working?

faheemhameed:
```
import "./App.css";
import { DataEditor, GridCellKind } from "@glideapps/glide-data-grid";

const columns = [
  { title: "First Name", width: 150 },
  { title: "Last Name", width: 150, textAlign: "right" },
];

function getData([col, row]) {
  const sampleData = "Hello, world!";

  return {
    kind: GridCellKind.Text,
    data: sampleData,
    displayData: sampleData,
    allowOverlay: false,
  };
}

function App() {
  return (
    <div className="App">
      <DataEditor getCellContent={getData} columns={columns} rows={10} />
    </div>
  );
}

export default App;
```

image


Here the Last Name column is supposed to be right-aligned  correct?

Please help.

jassmith:
The textAlign goes on the cell not the column

faheemhameed:
```
import "./App.css";
import { DataEditor, GridCellKind } from "@glideapps/glide-data-grid";

const columns = [
  { title: "First Name", width: 150 },
  { title: "Last Name", width: 150},
];

function getData([col, row]) {
  const sampleData = "Hello, world!";

  return {
    kind: GridCellKind.Text,
    data: sampleData,
    displayData: sampleData,
    allowOverlay: false,
    textAlign: "right",
  };
}

function App() {
  return (
    <div className="App">
      <DataEditor getCellContent={getData} columns={columns} rows={10} />
    </div>
  );
}

export default App;

```

It's the same alignment issue even though I set the textAlign on the Grid itself. (tried "right", "center", "end" etc)

I am not using TypeScript in my original project so I have tried it with plain javascript. I hope using plain js shouldn't be an issue right?

Please shed some light.

LukasMasuch:
@faheemhameed The alignment feature was implemented via the `contentAlign` keyword and not as initially planned via `textAlign`.

faheemhameed:
Thanks for that and it works very well. Lovely.

Is it also possible to apply contentAlign for the BaseGridColumn as well? IT doesn't seem to work on the columns as of now.
```

import "./App.css";
import { DataEditor, GridCellKind } from "@glideapps/glide-data-grid";

const columns = [
  { title: "First Name", width: 150 },
  { title: "Last Name", width: 150, textAlign: "center" },
];

function getData([col, row]) {
  const sampleData = "Hello, world!";

  return {
    kind: GridCellKind.Text,
    data: sampleData,
    displayData: sampleData,
    allowOverlay: false,
    contentAlign: "center",
  };
}

function App() {
  return (
    <div className="App">
      <DataEditor getCellContent={getData} columns={columns} rows={10} />
    </div>
  );
}

export default App;
```


image

jassmith:
Im not sure what the compelling argument for doing that would be other than
it being mildly simpler in cases where you want the whole column to work
that way.

LukasMasuch:
I think you are mainly referring to applying the alignment also to the column header, or? This sounds like a valid use case, but it could be more complicated to implement since the header can contain an icon/menu selector, etc... so we most likely would need to align all those elements.

jassmith:
Column header definitely has some issues that Im not sure how I want to
work out yet. I definitely dont want to be doing ellipsis in there.

faheemhameed:
Is there a workaround to just pass a custom component to the column and center the text in the custom component?

Thanks

jassmith:
There is a callback to override the render of the column header but then it's pretty much on you to center it correctly.

faheemhameed:
OK. Let me give it a try.

https://github.com/glideapps/glide-data-grid/blob/main/packages/core/API.md#gridcolumn

How can I use the callback, please? Can you please give me an example? Thanks.

PR #336, opened 2022-05-20
Fix vertical alignment in cells

LukasMasuch:
The vertical alignment of content in all (text-based) cells is currently always off by one pixel. I assume the reason is that subpixel rendering seems to not work with `fillText`. If the cell height is an even number, it will always be off by one pixel. In case the cell height is uneven it is possible to center the content. Therefore, I updated the default heights to uneven numbers.  However, we also need to apply `Math.ceil` here to get the correct centered y position. 

This PR currently fixes all cells besides:
- header (tried it but the correct way to center here is just using `ctx.fillText(c.title, drawX, y + height / 2);` which seems a bit strange)
- group header
- user-profile-cell
- sparkling cell
- range cell

### DIFF
--- a/packages/cells/src/cells/dropdown-cell.tsx
+++ b/packages/cells/src/cells/dropdown-cell.tsx
@@ -81,7 +81,7 @@ const renderer: CustomCellRenderer<DropdownCell> = {
         const { ctx, theme, rect } = args;
         const { value } = cell.data;
         ctx.fillStyle = theme.textDark;
-        ctx.fillText(value, rect.x + theme.cellHorizontalPadding, rect.y + rect.height / 2);
+        ctx.fillText(value, rect.x + theme.cellHorizontalPadding, rect.y + Math.ceil(rect.height / 2));
 
         return true;
     },
--- a/packages/cells/src/cells/range-cell.tsx
+++ b/packages/cells/src/cells/range-cell.tsx
@@ -34,7 +34,7 @@ const renderer: CustomCellRenderer<RangeCell> = {
         const { min, max, value, label, measureLabel } = cell.data;
 
         const x = rect.x + theme.cellHorizontalPadding;
-        const yMid = rect.y + rect.height / 2;
+        const yMid = rect.y + Math.ceil(rect.height / 2);
 
         const rangeSize = max - min;
         const fillRatio = (value - min) / rangeSize;
--- a/packages/cells/src/cells/star-cell.tsx
+++ b/packages/cells/src/cells/star-cell.tsx
@@ -83,7 +83,7 @@ const renderer: CustomCellRenderer<StarCell> = {
         drawX += 8;
         ctx.beginPath();
         for (let i = 0; i < stars; i++) {
-            pathStar(ctx, [drawX, rect.y + rect.height / 2], 16);
+            pathStar(ctx, [drawX, rect.y + Math.ceil(rect.height / 2)], 16);
             drawX += 18;
         }
         ctx.fillStyle = theme.textDark;
--- a/packages/cells/src/cells/tags-cell.tsx
+++ b/packages/cells/src/cells/tags-cell.tsx
@@ -91,14 +91,14 @@ const renderer: CustomCellRenderer<TagsCell> = {
 
         let x = drawArea.x;
         let row = 1;
-        let y = drawArea.y + (drawArea.height - rows * tagHeight - (rows - 1) * innerPad) / 2;
+        let y = drawArea.y + Math.ceil((drawArea.height - rows * tagHeight - (rows - 1) * innerPad) / 2);
         for (const tag of tags) {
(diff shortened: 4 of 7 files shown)


jassmith:
can you provide some before/after screenshots. We should also check alignment with the overlay editor to see if adjustments are needed there.

LukasMasuch:
Before:
_I will add some in few minutes._

After:
image
image
image

LukasMasuch:
I also added some CSS changes to fix the vertical alignment in overlay editor as well:

**Before:**
image
image

**After:**
image
image

LukasMasuch:
Do you know the reason for the `+1` here:

https://github.com/glideapps/glide-data-grid/blob/f3f3be6d0aeb4e06fce47d51c44f724ff525f348/packages/core/src/data-grid/data-grid-render.tsx#L685

jassmith:
I probably thought it looked better

LukasMasuch:
Btw. this change improves the vertical alignment, but it will still feel slightly misaligned depending on the font. The reason is that `textBaseline=middle` is not necessarily the middle of characters/numbers. A better way might be to use the `alphabetic` baseline and calculate an offset based on the height of capital letters of the used font. But I think this was changed in version `3.1.1` because of performance issues?

jassmith:
This was indeed changed in 3.1.1 due to performance issues. I think instead we could... and I know this is heathen speak, add a vertical offset into the theme.

jassmith:
The other way to do this is to make a font metric cache, and calculate the offset once and remember it.

dvdsgl:
This is fantastic.

LukasMasuch:
Yep, I think we could just measure one capital character once and cache this: 

```python
 ctx.measureText("H").actualBoundingBoxDescent
```

jassmith:
The cache would need to be nuked on the font load. I also consider a slight breaking change to change the default "baseline" for custom cells.

jassmith:
You also cant measure a single capital letter. Not all letters are the same cap height in many different fonts. It is safest to render A-Z and measure that. What I really am curious about is can we find some reasonable way to do this while keep baseline center. This would make custom cells ever so slightly easier and eliminate the back-compat break. I mean we could change it just when we go to render a custom cell, but I don't love the overhead. Changing the baseline has a pretty high synchronization cost.

jassmith:
We need a test bed.

jassmith:
So this is my best starting demo: https://codepen.io/jassmith/pen/VwQWBoz

It needs more testing with more fonts (I only tested 3), but I believe that this shows it might be possible to use baseline middle and still have an "ideal" centering we compute for all our internal font usage but leave easy centering available for custom cells and not break existing custom cells.

jassmith:
Nah I take it back, falls apart with Roboto. I just got lucky. I clearly don't understand the metrics like I hoped.

jassmith:
Okay the codepen is now updated with a fully working POC. The long and short of it is that you need metrics taken from textBaseline = "alphabetic" and textBaseline = "middle" and you can then reverse out the positions between the two. Shouldn't be too hard to write the algorithm if we have a cache that caches A-Z metrics for baseline and middle and is properly cleared on font load.

LukasMasuch:
I tried those snippets within glide-data-grid and I think both provide a very good vertical centering (at least for the default font):

**Alphabetic Baseline + Fix:**

```typescript
ctx.textBaseline = "alphabetic";

const metrics = ctx.measureText("ABCDEFGHIJKLMNOPQRSTUVWXYZ");
const fix = metrics.actualBoundingBoxAscent / 2;

const yCentered = y + Math.ceil(h / 2) + fix;
```

image

**Middle Baseline + Fix:**

```typescript
ctx.textBaseline = "alphabetic";
const metrics = ctx.measureText("ABCDEFGHIJKLMNOPQRSTUVWXYZ");

ctx.textBaseline = "middle";
const centeredMetrics = ctx.measureText("ABCDEFGHIJKLMNOPQRSTUVWXYZ");
const fix = -(centeredMetrics.actualBoundingBoxDescent - metrics.actualBoundingBoxDescent) + metrics.actualBoundingBoxAscent / 2;

const yCentered = y + Math.ceil(h / 2) + fix;
```

image


**Only Middle Baseline:**

```typescript
ctx.textBaseline = "middle";

const yCentered = y + Math.ceil(h / 2);
```

image

jassmith:
https://github.com/glideapps/glide-data-grid/pull/342 I've implemented the most basic version of this here. It probably needs to be applied to more cells and be exported and applied to all our custom cells.

jassmith:
@LukasMasuch would you mind reviewing the PR I posted to see if it meets your needs?

jassmith:
Lukas thank you so much for this excellent branch. It served as a basis for what has now landed in 4.0.4. I am closing this PR but it is in spirit already merged.

LukasMasuch:
Thanks for fixing the vertical alignment :)

PR #342, opened 2022-05-23
Add middle center bias algorithm

jassmith:


LukasMasuch (approved):
Awesome :) I have tested this version with a few tables/font styles I use and vertical centering seems to work great.

Nit: I think the actual center in all cells is still off by one pixel, which might be fixed by the  `Math.ceil` on the `height / 2` at least for cells with uneven height. But this is only a very subtle issue.

LukasMasuch:
I think the header / group header is still missing here: 

https://github.com/glideapps/glide-data-grid/blob/dd7033200b25bf3ac0015616cb7172dbca87c87f/packages/core/src/data-grid/data-grid-render.tsx#L521

https://github.com/glideapps/glide-data-grid/blob/dd7033200b25bf3ac0015616cb7172dbca87c87f/packages/core/src/data-grid/data-grid-render.tsx#L685

jassmith:
Good catch

PR #354, opened 2022-06-01
Allow drag and drop into cells

ivoelbert:
TODO:
- [ ] Add tests
- [ ] Document

This provides some low level primitives to handle drag and drop.

```ts
onDragOverCell?: (cell: Item, dataTransfer: DataTransfer | null) => void;
```

```ts
onDragLeave?: () => void;
```

```ts
onDrop?: (cell: Item, dataTransfer: DataTransfer | null) => void;
```

You can combine those two to add some nice drag and drop behavior.

The new added story uses `onDragOverCell` to add a single-cell highlight region if the dragged thing is supported, and `onDrop` to update the corresponding cell. `onDragLeave` is used to clean up the highlight region when leaving the grid.

https://user-images.githubusercontent.com/18490534/171917210-a4aeb086-33e6-49d3-8546-d62b67196e2b.mov

### DIFF
--- a/packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
@@ -2958,10 +2958,16 @@ export const DropEvents: React.VFC = () => {
                         },
                     },
                 ]);
+            } else {
+                setHighlights([]);
             }
         },
         [getCellContent]
     );
+
+    const onDragLeave = React.useCallback(() => {
+        setHighlights([]);
+    }, []);
 
     return (
         <BeautifulWrapper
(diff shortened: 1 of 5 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2730,6 +2732,8 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     onKeyDown={onKeyDown}
                     onMouseDown={onMouseDown}
                     onMouseUp={onMouseUp}
+                    onDragOverCell={onDragOverCell}
```
jassmith, line 2735: These need to deal with row marker row mangling

jassmith (changes requested):

PR #390, opened 2022-06-16
Updating API.md with latest trailingRowOptions

pzcfg:
The new options

https://github.com/glideapps/glide-data-grid/blob/91b2843ba5b248b7485293b8bc9f6fe53fb3a1fd/packages/core/src/data-grid/data-grid-types.ts#L182-L197

got updated in the GridColumn section

https://github.com/glideapps/glide-data-grid/blob/91b2843ba5b248b7485293b8bc9f6fe53fb3a1fd/packages/core/API.md?plain=1#L200-L220

but not in `trailingRowOptions`

### DIFF
--- a/packages/core/API.md
+++ b/packages/core/API.md
@@ -848,9 +848,11 @@ The grid will not attempt to add additional rows if more data is pasted then can
 
 ```ts
 trailingRowOptions?: {
-    readonly tint?: boolean;
     readonly hint?: string;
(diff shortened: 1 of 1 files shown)


jassmith:
Those API docs are for the global level trailingRowOptions, which also need updating. We need to update the docs for GridColumn likely too.

jassmith:
Fixed in 4.1.1 independently

PR #401, opened 2022-06-19
get rid of magic number in accessibilityTree

BrianHung:
Was looking through `data-grid` and noticed a magic constant that was introduced in this PR: https://github.com/glideapps/glide-data-grid/pull/121 and this commit: https://github.com/glideapps/glide-data-grid/commit/5ee2a7a0aaea31cd83fc5599f893612b38eda75d

Seems that we can replace it with `accessibilityHeight` but not exactly sure.


jassmith:
Nah this magic constant is more like "let's not waste time forcing react to render shit until the grid is likely near its final size."

It's definitely a hack but it works pretty well. The "width" in this case is just the pixel width of the grid. I guess technically that means any grid under 50px wide is not accessible but lots of things start breaking down at that size for this library.

PR #408, opened 2022-06-21
appendRow handle

pzcfg:
This adds `appendRow` to the `React.useImperativeHandle` so that it can be accessed via `ref.current.appendRow`

Also adds a story demonstrating its use

https://user-images.githubusercontent.com/94077014/174906326-e1be3bd7-98e4-445a-8593-d8d0af96c21e.mp4

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -277,6 +277,7 @@ export interface DataEditorProps extends Props {
 }
 
 export interface DataEditorRef {
+    appendRow: (col: number) => Promise<void>;
     updateCells: DataGridRef["damage"];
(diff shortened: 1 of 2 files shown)


jassmith:
I think there is an off by one issue due to row markers here.

pzcfg:
Aaah hmm, I think I can add a `+ rowMarkerOffset`, but if I add it to `appendRow` that's going to mess up how the grid adds internally, right? Should I add it in the `useImperativeHandle` directly? Something like?

```
React.useImperativeHandle(
        forwardedRef,
        () => ({
            appendRow: (col: number) => appendRow(col + rowMarkerOffset),
```

jassmith:
That looks right to me

PR #421, opened 2022-06-29
Improve scroll shadows

leolabs:
This PR improves shadows in the grid by making the opacity transition smoother, adding an optional shadow behind the header when scrolling, and making both shadows an option that can be turned off if needed.

https://user-images.githubusercontent.com/541628/176437926-64de5c52-97aa-46ab-a336-c166987ab50b.mp4


packages/core/src/data-grid/data-grid.tsx
```diff
@@ -47,6 +47,7 @@ import { AnimationManager, StepCallback } from "./animation-manager";
 import { browserIsFirefox } from "../common/browser-detect";
 import { CellRenderers } from "./cells";
 import { useAnimationQueue } from "./use-animation-queue";
+import { clamp } from "lodash";
```
jassmith, line 50: import clamp from "lodash/clamp";

It wont tree shake correctly like this.

packages/core/src/data-grid/data-grid.tsx
```diff
@@ -1313,23 +1325,75 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
     );
 
     const stickyShadow = React.useMemo(() => {
```
jassmith, line 1327: This memo now serves very little purpose. It will re-evaluate almost ever scroll step. If you compute `opacityX` and `opacityY` outside the memo and then pass those in it will only reevaluate when needed.

jassmith (changes requested):

jassmith (approved):

PR #423, opened 2022-06-30
Add allowOverlay description to API.md

Falieson:
add a little bit to the documentation to describe what allowOverlay does


jassmith:
Actually it means the editor will come up, but if `readonly` is set it may still not be editable

PR #436, opened 2022-07-13
Change column type from GridColumn to MappedGridColumn on DrawHeaderCallback

codeBelt:
I noticed that `DrawCustomCellCallback` `GridColumn` was missing the `sourceIndex` prop. I think this is the correct fix.

### DIFF
--- a/packages/core/API.md
+++ b/packages/core/API.md
@@ -204,6 +204,7 @@ Grid columns are the basic horizontal building block of the data grid. At their 
 ```ts
 interface BaseGridColumn {
     readonly title: string;
+    readonly sourceIndex: number;
     readonly group?: string;
     readonly icon?: GridColumnIcon | string;
     readonly overlayIcon?: GridColumnIcon | string;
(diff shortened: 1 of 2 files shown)


jassmith:
Why do you need the sourceIndex? You have the original columns you passed in, you should be able to get the index that way?

codeBelt:
When using `drawHeader` on `DataEditor` it looks like I only have access to the `DrawHeaderCallback` arg. Looking at the output I can access the column index by `column.sourceIndex` but it seems the type is wrong for `column` because it is missing `sourceIndex` and `sticky`. That is way I change the type from `GridColumn` to `MappedGridColumn`.

Screen Shot 2022-07-14 at 10 25 43 AM

I don't know of another way to get the column index when using the `drawHeader` prop.

jassmith:
I think I have an alternative solution which will work fro you.

jassmith:
I've pushed a change to add the index to the args instead. This has a similar effect except the index respect the rowMarkerOffset which means if you have rowMarkers enabled it will remain stable. The row marker column will have an index of -1.

codeBelt:
Ok, that works!

PR #453, opened 2022-07-30
Add capability for more granular drag sourcing without losing useful interactions

jassmith:


packages/core/src/data-editor/use-autoscroll.ts
```diff
@@ -0,0 +1,35 @@
...
+            } else {
+                const step = curTime - lastTime;
+                speedScalar.current = Math.min(1, speedScalar.current + step / msToFullSpeed);
+                const motion = speedScalar.current * step * maxPxPerMs;
+                scrollRef.current?.scrollBy(xDir * motion, yDir * motion);
+                lastTime = curTime;
+            }
+            t = window.requestAnimationFrame(scrollFn);
+        };
+        let t = window.requestAnimationFrame(scrollFn);
```
ivoelbert, line 32: Can't believe this just works and both TS and ESlint are fine with this.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1416,14 +1416,14 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
         [columns, getCellsForSelection, maxColumnWidth, mergedTheme, minColumnWidth, onColumnResize, rowMarkerOffset]
     );
 
+    const [scrollDir, setScrollDir] = React.useState<GridMouseEventArgs["scrollEdge"]>();
```
ivoelbert, line 1428: Why is this state? Can't we track this with a ref so the `onMouseMove` event (which will trigger a lot if you move the mouse) doesn't cause React redraws by default?

packages/core/src/data-editor/use-autoscroll.ts
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import type { GridMouseCellEventArgs } from "../data-grid/data-grid-types";
+
+const maxPxPerMs = 2;
+const msToFullSpeed = 3000;
+
+export function useAutoscroll(
```
ivoelbert, line 7: Now, if you make the scroll direction a ref instead of state, you'll have to change how this hook works. Perhaps an approach is to only loop rAFs if you are dragging (or if the mouse is pressed?), and read the ref on every loop to figure out how much you need to move.

packages/core/src/data-grid/data-grid.tsx
```diff
@@ -1000,13 +1013,21 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
     const onDragStartImpl = React.useCallback(
         (event: DragEvent) => {
             const canvas = ref.current;
-            if (canvas === null || !isDraggable) return false;
+            if (canvas === null || isDraggable === false || isResizing) {
+                event.preventDefault();
+                return false;
```
ivoelbert, line 1018: Why this `return false`? I don't think this does anything if you use `addEventListener`. Same for all the other ones.

ivoelbert (changes requested):

packages/core/src/data-editor/use-autoscroll.ts
jassmith: This one weird trick compiler engineers hate

packages/core/src/data-editor/data-editor.tsx
jassmith: Fixed

PR #455, opened 2022-08-02
check selection row is in range before updating

BrianHung:
Ran into bug where an invalid cell selection was being set when columns _and_ rows were decreased. Checking if col is in range is not sufficient because mangledCols can be stale in between useEffect calls that are dependent on cell selection only.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -2650,11 +2650,12 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
         const [col, row] = gridSelection.current.cell;
 
         // Check that the grid selection is in range before updating the selected cell
-        const selectionColInRange = mangledCols[col];
-        if (selectionColInRange === undefined) return;
+        const selectionColInRange = mangledCols[col] !== undefined;
+        const selectionRowInRange = row < mangledRows;
+        if (!selectionColInRange || !selectionRowInRange) return;
 
         updateSelectedCell(col, row, false, false);
-    }, [mangledCols, rows, gridSelection, updateSelectedCell]);
+    }, [mangledCols, mangledRows, gridSelection, updateSelectedCell]);
 
     const disabledRows = React.useMemo(() => {
         if (showTrailingBlankRow === true && trailingRowOptions?.tint === true) {


jassmith:
Please retarget PR to 5.0.0

jassmith:
Another problem, I no longer remember the point of this effect, can you help me?

BrianHung:
> Another problem, I no longer remember the point of this effect, can you help me?

Looks like it goes back to the initial commit: https://github.com/glideapps/glide-data-grid/blame/77adf343208dd90a1b361f71a1922e8cbc7c0d11/src/data-editor/data-editor.tsx#L922-L926 Not exactly sure without testing why you couldn't remove it 🤔

jassmith:
I *think* its purpose is to re-trigger the scrolling logic so the selected cell stays on screen. Testing will confirm.

BrianHung:
Rebased onto `5.0.0`.

jassmith:
I do not believe this effect does anything anymore. Simply deleting it seems to be the prudent path forward.

jassmith:
Ok I dug into our internal repo which contains the original product the data editor was forked out of. This is the commit message for that effect: "Ensure data editor cannot select out of bounds cells"

So the question is, do we still have an issue with this?

BrianHung:
> I do not believe this effect does anything anymore.

It still does indirectly keep the cell in view via `updateSelectedCell` when its controlled externally via `gridSelection` (unless there's another thing keep it in view?).

> So the question is, do we still have an issue with this?

Think that depends on whether glide should fix out of bound cell selections given via props, or whether it should leave it up to end developer to take care of that. For example, what should happen when a row or column is deleted?

On my end, this PR fixed an issue caused by 
1. `updateSelectedCell` where rows and columns were both decreased, but this useEffect hook which is dependent on rows / mangledRows ran before my useEffect hook that computes columns data and is dependent on column count.
2. `setGridSelection` was called from inside of `updateSelectedCell` with the old cell selection that is valid with old columns (invalid with new rows),
3. my useEffect column hook ran to modify columns, 
4. and then a useEffect hook on `cell` that uses `getCellBounds` encountered invalid width property on `columns`.

jassmith:
> It still does indirectly keep the cell in view via updateSelectedCell when its controlled externally via gridSelection (unless there's another thing keep it in view?).

I can't reproduce this problem in testing after deleting the effect, can you? As for should it coerce the gridSelection, ideally yes. It should rewrite the gridSelection to not include out of range areas.

jassmith:
I really need a good reproduction case, do you think you could get one in this PR?

BrianHung:
> EDIT - I can reproduce it not going to the selected cell when its updated externally, but it did that with or without the effect.

Huh, what's keeping the cell selection in view when updated externally then? Pretty useful to save end developer from having to call `dataGridRef.scrollIntoView` after each `setGridSelection`.

> I really need a good reproduction case, do you think you could get one in this PR?

I'll try my best to replicate 👍

jassmith:
Nothing is as best I can tell. I could not replicate it keeping things in view when the gridSelection is modified externally. I am not denying this would be a useful behavior however. I will try to make it happen.

jassmith:
I've pushed an update which I believe fixes all the issues including the crasher you had. Instead of crashing when requesting out of bounds cells it will now return an empty rectangle.

PR #467, opened 2022-08-12
Pass fixedShadowX and fixedShadowY down to child components

ralake:
Thee props aren't being passed down into child components so setting them to false right now when rendering DataGrid doesn't do anything.


jassmith:
I already fixed this in 5.0.1 (unreleased) :) Thank you for the PR!

PR #486, opened 2022-08-29
5.1.0

jassmith:
# 5.1.0 Release Notes

## New API
- `Theme` now exports `headerBottomBorderColor` to control the color of the line directly under the header.
- `onKeyUp`/`onKeyDown` are now available to end users
- `drawFocusRing` can be set to false to disable the drawing of the focus ring.
- Custom `ImageWindowLoader` can be passed to the `DataEditor` to allow for custom image loading handling.
- `customRenderer` API - FIXME
- `onColumnResize` now has a 4th parameter which reports the new size of the column including any pixels added by the column grow parameter being set.
- `onScroll` API now supports scrolling to a specific pixel value and specifying the scroll target behavior
- `ImageCell` can now specify the rounding radius of the image being drawn
- Declarative model for setting scroll offset via `scrollOffsetX` and `scrollOffsetY`
- Add ability to control max auto column width independently of max column width settable by users
- initialScrollOffset

## Improvements
- Our monorepo is now a fairly standard monorepo setup, huzzah
- Links cell added to `glide-data-grid-cells`
- Button cell added to `glide-data-grid-cells`
- Code splitting now works correctly again
- Bounce scrolling now supported on iOS and MacOS
- Add ability to set rendering strategy in experimental flags

## Bug Fixes
- Fixed a rare bug that can cause flickering lines
- Glide Data Grid now properly handles inputs when used in a scaled environment
- Scrolling to the selected cell now works correctly when using a controlled `gridSelection`
- DropdownCell no longer automatically selects the first item and instead selects the current value.


package.json
```diff
@@ -77,5 +83,7 @@
         "typescript": "^4.5.2",
         "webpack": "^5.73.0"
     },
-    "dependencies": {}
+    "resolutions": {
```
jassmith, line 86: Remove me

packages/cells/tsconfig.json
```diff
@@ -1,16 +1,5 @@
...
-        "strict": true,
-        "moduleResolution": "node",
-        "target": "es6",
-        "jsx": "react",
-        "typeRoots": ["./node_modules/@types"]
-    },
-    "include": ["./src/**/*.ts", "./src/**/*.tsx"]
+    "extends": "../../tsconfig.json",
+    "include": ["./src/**/*.ts", "./src/**/*.tsx"],
+    "references": [{ "path": "../core" }]
```
jassmith, line 4: is this working?

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1486,6 +1604,8 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 return false;
             };
 
+            if (isPrevented.current) return;
```
jassmith, line 1607: Review if needed

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1994,6 +2116,19 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
     const onKeyDown = React.useCallback(
         (event: GridKeyEventArgs) => {
             const fn = async () => {
+                let cancelled = false;
+                if (onKeyDownIn !== undefined) {
+                    onKeyDownIn({
+                        ...event,
+                        cancel: () => {
+                            cancelled = true;
+                            event.cancel();
```
jassmith, line 2125: need to figure out if this is actually desired, probably not. This also calls `stopPropagation`

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2655,7 +2761,12 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
     const scrollToRef = React.useRef(scrollTo);
     scrollToRef.current = scrollTo;
     React.useEffect(() => {
-        if (outCol !== undefined && outRow !== undefined) {
+        if (
+            outCol !== undefined &&
+            outRow !== undefined &&
+            (outCol !== expectedExternalGridSelection.current?.current?.cell[0] ||
```
jassmith, line 2767: This is probably too naughty of a hack. The better solution seemingly is to not pre-scroll when the user is controlling the gridSelection and always use this effect to perform any selection maintenance scrolling.

packages/core/src/data-grid-overlay-editor/private/markdown-overlay-editor.tsx
```diff
@@ -19,18 +18,10 @@ interface Props {
     createNode?: (content: string) => DocumentFragment;
 }
 
+// TODO: Fix enter key behavior
```
jassmith, line 21: Remove me

packages/core/src/data-grid/cells/index.ts
```diff
@@ -15,22 +14,22 @@ import { textCellRenderer } from "./text-cell";
 import { uriCellRenderer } from "./uri-cell";
 
 // we are giving up type safety here but I cant figure out right now how to do this right.
-const asCollapsed = (x: any) => x as InternalCellRenderer<InnerGridCell>;
+// const asCollapsed = (x: any) => x as InternalCellRenderer<InnerGridCell>;
```
jassmith, line 17: Remove me

packages/core/src/data-grid/cells/index.ts
```diff
@@ -15,22 +14,22 @@ import { textCellRenderer } from "./text-cell";
...
 
-type RendererKinds = InnerGridCellKind | Exclude<GridCellKind, GridCellKind.Custom>;
+// type RendererKinds = InnerGridCellKind | Exclude<GridCellKind, GridCellKind.Custom>;
```
jassmith, line 19: Remove me

packages/core/src/data-editor/data-editor.tsx
jassmith: I've opted to keep this for now but have added some extra testing.

LukasMasuch:
Thanks again for this awesome update :) I tried out the `5.0.1-alpha2` and saw a small issue with the cell borders in the first row: 
image

jassmith:
awwwww nuts I knew there was going to be some bug due to that...

jassmith:
The coveralls change is unexpected and needs explanation

LukasMasuch:
Something changed between `alpha3` and `alpha4` impacting resizing of the dataframe:

`alpha3`:

https://user-images.githubusercontent.com/2852129/188611920-f70213af-b2be-4963-85e7-61f5ec435aad.mov


`alpha4`:

https://user-images.githubusercontent.com/2852129/188611937-30522deb-26c7-474a-9dd5-0d842e1e8509.mov

jassmith:
Can I have you testcase please? I cannot reproduce this regression.

jassmith:
This is the diff between alpha3 and alpha4: https://github.com/glideapps/glide-data-grid/compare/aa9c6149f65338ed9bd1351421c58f44f20c1c8c...bc3e0cab190ac01b8f5f84fddfc24ad57185426d

I cannot for the life of me see what would account for this

jassmith:
image

This screen shot from your video shows the canvas being out of sync with its rendered size. This was a bug we had in the past but I can't imagine how its back...

LukasMasuch:
Just tested it again and it consistently happens in alpha4 and not alpha3 (just yarn upgrade between both versions and leave everything the same). But it only happens if the data grid is used with a resizable container based on [re-reziable](https://github.com/bokuweb/re-resizable) and it works fine with the native CSS resize. I did have a quick look through the differences, but it seems that nothing should have impacted anything related to this. 

Unfortunately, I won't be able to create a separated testcase today, but will try to provide something reproducible later this week.

jassmith:
@LukasMasuch I really need a reproduction case or Im not going to help up the release for this :(

jassmith:
I made a demo here, it does not show the issue you are having. https://github.com/glideapps/glide-data-grid/tree/re-resizer

LukasMasuch:
Sorry for the delay. I'm on vacation and do not have my work computer with me. However, I managed to create a reproducible example based on your demo with the following snippet:

```typescript
const GridContainer = styled.div`
    position: relative;

    & .glideDataEditor {
        min-width: 100%;
    }
`;

export const Resizer: React.VFC = () => {
    const { cols, getCellContent } = useMockDataGenerator(1000, true, true);

    const [[width, height], setSize] = React.useState([100, 100]);

    return (
        <BeautifulWrapper title="Resizer" description={undefined}>
            <GridContainer>
                <Resizable
                    defaultSize={{ width, height }}
                    onResize={(_e, _dir, _el, delta) => setSize([width + delta.width, height + delta.height])}>
                    <DataEditor
                        className="glideDataEditor"
                        {...defaultProps}
                        width={width}
                        height={height}
                        getCellContent={getCellContent}
                        columns={cols}
                        rows={3000}
                        rowMarkers="both"
                    />
                </Resizable>
            </GridContainer>
        </BeautifulWrapper>
    );
};
(Resizer as any).parameters = {
    options: {
        showPanel: false,
    },
};
```

https://user-images.githubusercontent.com/2852129/189332417-a2ad6f04-f665-4d7f-ad2b-683966d3f789.mov


These are only the minimal extracted parts to reproduce the issue. In my actual setup, the `GridContainer` adapts to the minimal size of the table based on the calculated column widths, and the `Resizable` adapts to `100%` of the parent (==  `GridContainer`) width. I'm using an additional container because `re-reziable` does not provide the capability to adapt the initial size to the dimensions of the children (`DataEditor`). This worked fine with `alpha3`, but not anymore with `alpha4`.

LukasMasuch:
btw. I used Codespace to test this and`Contribution` docs seem to be not yet updated to the changes in this branch.`npm run install`  fails since `install` does not exist as a script. Probably it is referring to `npm install`. Hower, in addition to running `npm install` in the root, I also had to run it in all the packages (`core`, `cells`, and `source`) to get the storybook working.

LukasMasuch:
I figured out the reason for the issue. The changes in https://github.com/glideapps/glide-data-grid/compare/aa9c6149f65338ed9bd1351421c58f44f20c1c8c...bc3e0cab190ac01b8f5f84fddfc24ad57185426d add the specified `className` also to a lot of the underlying components (e.g. `dvn-scroller` and `canvas`). This causes the issue with my setup. Knowing this, it is easy to fix for me. But not sure if applying the `className` to many of the underlying components is a wanted behavior?

jassmith:
Woops that isn't on purpose, thank you for catching that

jassmith:
Classname issue resolved.

PR #517, opened 2022-09-30
Drop a couple sig figs

Tonlsv:


jassmith:
What the heck is this?

PR #555, opened 2022-11-17
Improve documentation

tmcw:
Hi! This is an incremental improvement to Glide's docs that I've been writing as I use it, to try and clear up some confusion, improve inter-linking docs, and make it possible to implement the icon scheme by reading the docs.

### DIFF
--- a/packages/core/src/data-grid/data-grid-types.ts
+++ b/packages/core/src/data-grid/data-grid-types.ts
@@ -240,7 +240,12 @@ export type CellArray = readonly (readonly GridCell[])[];
 
 /**
  * This type is used to specify the coordinates of
- * a cell within the dataset.
+ * a cell or header within the dataset: positive row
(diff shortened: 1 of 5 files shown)


packages/core/src/data-grid/data-grid-types.ts
```diff
@@ -238,7 +238,12 @@ export enum GridColumnIcon {
 /** @category Types */
 export type CellArray = readonly (readonly GridCell[])[];
 
-/** @category Types */
+/**
+ * This type is used to specify the coordinates of
+ * a cell within the dataset.
```
jassmith, line 243: a cell or header

jassmith: headers are at row -1, group header at row -2

jassmith (approved):

PR #556, opened 2022-11-17
An attempt at in-memory undo/redo support

aaronSig:
This add a hook & storybook example showing in-memory undo and redo support. It works by wrapping the `getCellContent` and `setCellValue` functions so it can fetch a cell before it is updated and replay any updates back. 

Open to feedback and no pressure to merge if you don't think this is in scope. It may be better done in the datasource.

### DIFF
--- a/packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
+++ b/packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
@@ -42,6 +42,7 @@ import {
 } from "./utils";
 import noop from "lodash/noop.js";
 import type { GetRowThemeCallback } from "../../data-grid/data-grid-render";
+import { useUndoRedo } from "../use-undo-redo";
 
 export default {
     title: "Glide-Data-Grid/DataEditor Demos",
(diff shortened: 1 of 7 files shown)


packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
```diff
@@ -3612,3 +3613,54 @@ export const DropEvents: React.VFC = () => {
...
+                getCellContent={getCellContent}
+                gridSelection={gridSelection ?? undefined}
+                onGridSelectionChange={onGridSelectionChange}
+                columns={cols}
+                rows={1000}
+            />
+        </BeautifulWrapper>
+    );
+};
+(UnevenRows as any).parameters = {
```
jassmith, line 3662: Wrong story :)

jassmith (changes requested):
You are right about moving this to the source package. Other than that this looks fine to me. We can work on improvements incrementally.

packages/core/src/data-editor/stories/data-editor-beautiful.stories.tsx
aaronSig: Ah - thanks for catching.

aaronSig:
That's great! 

I've fixed up that typo & moved the hook down into the source package. I needed some of the utils for the story so I've moved those as well.

jassmith (approved):

tortarino:
Hello,

Thanks to both of you for all the great work!

It seems the undo/redo feature has not made it to the npm repository:
https://www.npmjs.com/package/@glideapps/glide-data-grid-source?activeTab=code

Am I missing something?

Thanks

aaronSig:
No, it doesn't look like it's in npm. 

The undo/redo hook is all self contained in one file though `packages/source/src/use-undo-redo.ts`

You could get by just copying that into your project.

PR #589, opened 2023-01-11
disable col resize cursor on row marker col edge

BrianHung:


BrianHung:
@jassmith Oh wait it might be `firstColAccessible ? 1 : 0`. There's not a direct `rowMarkerOffset` passed as a prop, but if there's a row marker, it should be `location[0] >= 1`, else `>= 0`

jassmith:
I haven't tested yet, but I believe the way you have it is right from memory. If you had it the wrong way round I was just going to reverse it for you.

PR #590, opened 2023-01-11
fix scrollTo when canvas is scaled

BrianHung:


jassmith:
Failing test :)

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1367,6 +1367,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     }
 
                     const scrollBounds = canvas.getBoundingClientRect();
+                    const scale = scrollBounds.width / canvas.offsetWidth;
```
jassmith, line 1370: Does this work right when there is a physical scrollbar?

BrianHung: What do you mean by physical scroll bar?

Does physical scroll bar width affect canvas width and scale somehow?

jassmith: I think if there is a scrollbar (like not an overlay one, a permanent one) the scrollBounds and the canvas bounds will have different widths when unscaled.

BrianHung: Testing this on a windows computer, the scaling factor computed above is almost exact to what I set via transform (some trailing decimal points that are minor).

jassmith: 👍 will test myself thank you

PR #627, opened 2023-02-01
DatePickerCell Implementation

willhuang1997:
Hi,

I am a software engineer at Streamlit working with Lukas Masuch. I am attempting to add date, datetime, and time support but it looks like the current implementation doesn't fully work and there are some edge cases that aren't necessary kept. 

I left comments as to why most changes are specifically there. I didn't necessarily see any tests that would be implemented but I have tested this custom cell in Streamlit, in glide-data-grid, and also had Lukas look over some of the code as a review (obviously doesn't count as a review but just saying). 

I would love to work with anyone who would like to help me get this in and any requests to make this work! 

William Huang
SWE at Streamlit / Snowflake


jassmith:
Most polite PR ever

jassmith:
Did you run prettier over these files? The formatting feels off. I should have a check for that...

willhuang1997:
> Did you run prettier over these files? The formatting feels off. I should have a check for that...

Hm... I ran:

`yarn prettier --write packages/` 

and then committed...?

jassmith:
No problem

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,87 @@
...
+    GridCellKind,
+    ProvideEditorCallback,
+} from "@glideapps/glide-data-grid";
 
 interface DatePickerCellProps {
     readonly kind: "date-picker-cell";
     readonly date: Date | undefined;
     readonly displayDate: string;
-    readonly format: "date" | "datetime-local";
+    readonly format: "date" | "time" | "datetime-local";
```
jassmith, line 14: extract to private type

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,87 @@
...
 }
 
+export const formatValueForHTMLInput = (format: "date" | "time" | "datetime-local", date: Date): string => {
+    if (format === "date") {
```
jassmith, line 18: switch (format) {
  case "data": // blah blah
  ...
  default: assertNever(format);

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,87 @@
...
+            required
+            style={{ minHeight: 26, border: "none", outline: "none" }}
+            type={format}
+            value={value}
+            autoFocus={true}
+            onChange={event => {
+                // handle when clear is clicked and value has been wiped
+                if (event.target.value === "") {
+                    try {
+                        cell.onChange({
```
jassmith, line 52: What are the circumstances you expect for this try to throw?

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,87 @@
...
+                    }
+                    return;
+                }
+                cell.onChange({
+                    ...cell.value,
+                    data: {
+                        ...cell.value.data,
+                        // use valueAsNumber because valueAsDate is null for "datetime-local"
+                        // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#technical_summary
+                        date: new Date(event.target.valueAsNumber) ?? cellDataDate,
```
jassmith, line 77: Can we get some unit tests around some of this behavior? I know that might be hard but I would appreciate some ideas of what you think should be tested.

packages/cells/src/cells/date-picker-cell.tsx
willhuang1997: in the case the displayDate is some weird string somehow.

packages/cells/src/cells/date-picker-cell.tsx
jassmith: never too late to add `data-picker-cell.test.tsx` 👍 

The cells package is definitely lacking on the tests front.

willhuang1997: Cool I will do that!

packages/cells/src/cells/date-picker-cell.tsx
jassmith: Can you point me to the code you think would throw? Perhaps the smell here is that we should be protecting at the far end.

jassmith: 👍

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,87 @@
...
+                    } catch (error) {
+                        cell.onChange({
+                            ...cell.value,
+                            data: {
+                                ...cell.value.data,
+                                displayDate: String(error),
+                            },
+                        });
+                    }
+                    return;
```
jassmith, line 69: I find this early return hurts readability in this case. The error case here is kind of "the bigger chunk" of the method. Can we move this to just a regular if/else setup for now?

willhuang1997: Are you suggesting this:
```
// handle when clear is clicked and value has been wiped
if (event.target.value === "") {
    cell.onChange({
      ...cell.value,
      data: {
        ...cell.value.data,
        // attempt to reset to cached date
        date: date !== undefined ? date : new Date(displayDate),
      },
    });
  } else {
    cell.onChange({
      ...cell.value,
      data: {
        ...cell.value.data,
        // use valueAsNumber because valueAsDate is null for "datetime-local"
        // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#technical_summary
        date: new Date(event.target.valueAsNumber) ?? cellDataDate,
      },
    });
  }
                        ```

Comment (jassmith) [reply packages/cells/src/cells/date-picker-cell.tsx:69]: yes

Comment (jassmith) [reply packages/cells/src/cells/date-picker-cell.tsx:69]: well kind, you removed the try catch in that block, but essentially yes

File: packages/cells/src/cells/date-picker-cell.tsx (hunk 4 above)
Comment (willhuang1997) [reply packages/cells/src/cells/date-picker-cell.tsx:77]: This might be a dumb question but where exactly would the unit test go? I created a date-picker-cell.test.tsx in the cells package. However, when I run a `npm run test`, it's only running the core tests. As a result, what would be the best way organizationally to add this test?

Comment (jassmith) [reply packages/cells/src/cells/date-picker-cell.tsx:77]: Oh uh... heh... oh what a day to be called out as a software developer. Yeah... you would need to add the jest command and setup and ya know what I will add the framework and you can rebase onto the 5.2.2 branch as soon as I add that...

File: packages/cells/src/cells/date-picker-cell.tsx (hunk 1 above)
Comment (willhuang1997) [reply packages/cells/src/cells/date-picker-cell.tsx:14]: Can I rename this to type? I think type is a better variable name as maybe I want to add a way to pass in formatting for a future PR. 

Assuming you're using moment, it would be along the lines of:
`moment(date).format(format)`

Could use any library but just for the current moment's sake... :) 

And for formatting support, I would likely make a PR for that too probably.

Comment (jassmith) [reply packages/cells/src/cells/date-picker-cell.tsx]: rename it to `dateKind` as it is the kind of date in the `date` param

File: packages/cells/src/cells/date-picker-cell.tsx [hunk 6]
```diff
@@ -1,15 +1,80 @@
-import * as React from "react";
-import { CustomCell, CustomRenderer, drawTextCell, GridCellKind } from "@glideapps/glide-data-grid";
+import React from "react";
+import { assertNever } from "../../../core/src/common/support";
```
willhuang1997, line 2: Is this the way to do this?

packages/cells/src/cells/date-picker-cell.tsx
jassmith: You will need to support both initially however. We will mark format as deprecated. When you come back and want to add format support we will make the full transition.

packages/cells/src/cells/date-picker-cell.tsx
jassmith: fuck...

jassmith: duplicate that method into the cells package.

willhuang1997: just the method or the whole support.ts? For now, I will do just that method :)

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,80 @@
...
+        <input
+            required
+            style={{ minHeight: 26, border: "none", outline: "none" }}
+            type={dateKind}
+            value={value}
+            autoFocus={true}
+            onChange={event => {
+                // handle when clear is clicked and value has been wiped
+                if (event.target.value === "") {
+                    cell.onChange({
```
jassmith, line 54: Why did you get rid of the try catch now?

jassmith (changes requested):

packages/cells/src/cells/date-picker-cell.tsx
willhuang1997: Added it back! This is what we're looking for?

willhuang1997: Also, I will have to head out soon for a good amount of time but I might come back to this later tonight / tomorrow morning :) so if I stop replying / commiting, sorry but we can continue later :)
Thanks for the fast replies and commentary.

jassmith: Hey no problem, looks great so far!

packages/cells/src/cells/date-picker-cell.tsx
jassmith: 👍

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,90 @@
...
 
 interface DatePickerCellProps {
     readonly kind: "date-picker-cell";
     readonly date: Date | undefined;
     readonly displayDate: string;
-    readonly format: "date" | "datetime-local";
+    readonly dateKind: DateType;
 }
 
+export type DateType = "date" | "time" | "datetime-local";
```
jassmith, line 18: rename to `DateKind`

jassmith (approved):
Everything else looks great. Will try to get test support stood up ASAP so Im not blocking you.

jassmith:
5.2.2 now has a test framework in place for cells. Please rebase your branch onto 5.2.2 and retarget.

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,80 @@
...
+            autoFocus={true}
+            onChange={event => {
+                // handle when clear is clicked and value has been wiped
+                if (event.target.value === "") {
+                    cell.onChange({
+                        ...cell.value,
+                        data: {
+                            ...cell.value.data,
+                            // attempt to reset to cached date
+                            date: date !== undefined ? date : new Date(displayDate),
```
LukasMasuch: One question here would be if we even want to have this behavior. I think we have the following options in case the user selects clear in the date picker dialog:
1) reset the value (`date`) to the current value of the cell (this is what the implementation from above is doing)
2) set the value (`date`) to `undefined` and thereby clearing the cell. 

I think both options seem to be reasonable, but from playing around a bit with the input components, I have a slight preference for `2` (clearing the cell value).

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,80 @@
...
+
 export type DatePickerCell = CustomCell<DatePickerCellProps>;
 
+const Editor: ReturnType<ProvideEditorCallback<DatePickerCell>> = cell => {
+    const cellData = cell.value.data;
+    const { date, displayDate, format } = cellData;
+    const cellDataDate =
+        cellData.date ??
+        // now as date with adjusted timezone to allow for displayDate to be correct
+        new Date(new Date().getTime() - new Date().getTimezoneOffset() * 60000);
```
LukasMasuch, line 41: I think there are also two options here in case cell value is `undefined` and user opens the editor:
1) Initializing the value of the input with the user's current date/time (this implementation)
2) Setting the value of the input to empty/undefined -> this would show the empty template of the input, e.g.:
image

My preference would be option 2 here since it is a bit more consistent with other existing cell editors showing the actual current value of the cell.

packages/cells/src/cells/date-picker-cell.tsx
willhuang1997: This will unfortunately set the underlying date to undefined and make the cell null. If the cell has a value and then you do a clear, I would expect as a user that it just clears the calendar and not the actual cell, especially since it doesn't close the calendar / editor.

packages/cells/src/cells/date-picker-cell.tsx
willhuang1997: That's a good point. I'll change it to undefined.

packages/cells/src/cells/date-picker-cell.tsx
jassmith: Thanks Lukas :)

packages/cells/src/cells/date-picker-cell.tsx
jassmith: Thanks Lukas :)

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,15 +1,89 @@
...
+import { assertNever } from "./common/support";
+import {
+    CustomCell,
+    CustomRenderer,
+    drawTextCell,
+    GridCellKind,
+    ProvideEditorCallback,
+} from "@glideapps/glide-data-grid";
 
 interface DatePickerCellProps {
```
LukasMasuch, line 11: I think it would be awesome if we also support the `readonly` property for this cell. So, basically we add `readonly readonly?: boolean;` here and either 1) just set the input field to [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)  or 2) return a disabled [GrowingEntry](https://github.com/glideapps/glide-data-grid/blob/b470dee3845a817b32ad36d4d96c21f47547211e/packages/core/src/growing-entry/growing-entry.tsx#L16) (exposed as TextCellEntry) containing the `displayDate` (my preferred solution)

LukasMasuch:
Another more optional feature (does not need to be part of this PR): Make the `min` (string), `max` (string), and `step`  (number) attributes accessible through the cell props. These parameters are supported in the same way by the time, date, and datetime-local inputs, so I think this should be very trivial to support.

packages/cells/src/cells/date-picker-cell.tsx
jassmith: I like option 2 better

willhuang1997: That's in the works! I like options 2 too! :)

willhuang1997: @jassmith , for the growingEntry, that's in the core package. Should I just copy over that package to the cells package?

jassmith:
I probably wont have time for about 7 days to debug what is going on there. If you can get it working I will have time to review however.

willhuang1997:
Hi @jassmith , I added some tests but I couldn't quite get the onChange tests implemented. However, I manually tested it. Could I still get a review or do I need the onChange tests?

packages/cells/test/date-picker-cell.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    function getMockEditorTarget(): Rectangle {
+        return {
+            x: 0,
+            y: 0,
+            width: 100,
+            height: 32,
+        };
+    }
+
+    it('renders into the dom with correct value', async () => {
```
LukasMasuch, line 44: Maybe we can add a test here for `date`, `time`, and `datetime` kind.

willhuang1997: done

packages/cells/test/date-picker-cell.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        const input = result.getByTestId("test-id")
+        expect(input).not.toBeUndefined()
+
+        // @ts-ignore
+        expect(result.value === "04:47:44.584");
+    });
+})
+
+describe("onPaste", () => {
+    it('correctly returns a value when onPaste is called with valid value', () => {
```
LukasMasuch: Since the string value that is passed in can be anything, it would be good to add more cases here (e.g. via `it.each`). `new Date` (or the equivalent `Date.parse`) support a few more date formats as you can see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#several_ways_to_create_a_date_object) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#date_time_string_format). This should include also string with only a date. Since the user theoretically can paste in anything, it would be good to add some more invalid examples as well. For example, a correct ISO date string but with a wrong range: e.g. `2023-20-21`

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -1,68 +1,126 @@
...
+    (cell.data as any).kind === "date-picker-cell",
+  draw: (args, cell) => {
+    const { displayDate } = cell.data;
+    drawTextCell(args, displayDate, cell.contentAlign);
+    return true;
+  },
+  provideEditor: () => ({
+    editor: Editor,
+  }),
+  onPaste: (v, d) => {
```
LukasMasuch, line 112: Maybe this should support time-only formats as well in `onPaste`. Without using another library, the best option might be to try converting the pasted value into a full date string (in case `newDate` is Nan/undefined):

```
new Date(`1970-01-01T${v}Z`)
```

willhuang1997:
@jassmith , I have added all of the tests in order to test all functionality. PTAL when you have time :)

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -18,46 +101,20 @@ const renderer: CustomRenderer<DatePickerCell> = {
...
             /* do nothing */
         }
-
+        if (Number.isNaN(newDate)) {
+            // Try to interpret value as time string (HH:mm:ss)
+            newDate = new Date(`1970-01-01T${v}Z`);
+        }
         return {
             ...d,
             date: Number.isNaN(newDate) ? undefined : newDate,
```
LukasMasuch, line 120: Oh, I think `Number.isNaN` is actually not working correctly here. For example,  `new Date("foo")` will not be detected via `Number.isNaN`. This only works in combination with `Date.parse`. There are a few information [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#fall-back_to_implementation-specific_date_formats).

LukasMasuch: I will add a fix

willhuang1997: I think this could be `Number.isNaN(newDate.getTime())`

LukasMasuch: yep, but I think it's better to directly use `Date.parse`, since this is anyways used when called via `new Date()`

willhuang1997:
Hi @jassmith ! I know you're busy but will you have time to review it anytime soon? Is it possible to get a date as to when this will be reviewed? 

Thanks,

William

jassmith:
I will review it tonight after dinner

packages/cells/src/cells/date-picker-cell.tsx
```diff
@@ -18,49 +115,29 @@ const renderer: CustomRenderer<DatePickerCell> = {
...
-                val = val.split("T")[0];
-            } else {
-                val = val.substring(0, 23);
+            if (Number.isNaN(parseDateTimestamp)) {
+                // Support for parsing ISO 8601 date strings:
+                parseDateTimestamp = Date.parse(v);
+                if (d.format === "time" && Number.isNaN(parseDateTimestamp)) {
+                    // The pasted value was not a valid date string
+                    // Try to interpret value as time string instead (HH:mm:ss)
+                    parseDateTimestamp = Date.parse(`1970-01-01T${v}Z`);
```
jassmith, line 134: Why not interpret it as "today" instead of 1970? Today at least feels more likely to be expected.

jassmith (approved):
Looks good overall. I am curious about the 1970 choice.

willhuang1997:
> Looks good overall. I am curious about the 1970 choice.

Thanks! I think any day can be consistent but 1970 is just because it's the Unix epoch. Is it possible to merge it? Only those with write access can merge :)

packages/cells/src/cells/date-picker-cell.tsx
LukasMasuch: I assume we could also use today's date here...  both are probably semi-optimal solutions here. The general problem here is that there isn't really a native format for time in javascript.

LukasMasuch: thinking a bit more here. Probably today has slightly more advantages 👍 @willhuang1997 would you be fine as well with changing that?

LukasMasuch:
Oh, I think the tests are not yet running as part of the CI 😬 Let me quickly add that

packages/cells/src/cells/date-picker-cell.tsx
LukasMasuch: One argument for keeping `1970-01-01`: It is consistent with how the time input converts the time back into a date. Calling `valueAsDate` on the time input event will return the time based on `1970-01-01`. `valueAsNumber` is also the UNIX timestamp from `1970-01-01` and not today.

image

jassmith:
Thank you both for your excellent work. I got this one from here. This looks great.

willhuang1997:
Since the build was broken, I removed the failing tests because they were based off of local time zone and causing flakiness.

LukasMasuch:
It would be great if we could make the DatePickerCell to also adapt to the content when it's automatically sized. We would need to add this to the `renderer`:

```
  measure: (ctx, cell) => {
    const { displayDate } = cell.data
    return ctx.measureText(displayDate).width + 16
  },
```

But in order to have that work, we need to remove this check (see the issue [here](https://github.com/glideapps/glide-data-grid/issues/648)): 

https://github.com/glideapps/glide-data-grid/blob/c9e1f3df5d545171ccca2226c0c55e253fb6cc83/packages/core/src/data-editor/use-column-sizer.ts#L24

LukasMasuch:
I added a couple of small refinements and improvements to the PR:
- Add support for `null` as date value.
- Use `defaultValue` instead of `value`. Using `value` can lead to some strange issues. 
- Check if `valueAsNumber` is NaN instead of empty `value` to determine that the created value by the user is not valid since we are using `valueAsNumber` and not `value`.
- Add `measure` to allow auto-resizing of the cell.
- Allow providing a `timezoneOffset`. This allows to let the user to edit a datetime object in the date & time of the timezone, but still get back the actual UTC date adjusted by the timezone.
- Adjust paste tests to only use strings, since pasting will always use strings anyways.

msdrigg:
I'd love to see this merged in. It looks like a great set of features.

msdrigg:
Is anything blocking this? This feature set looks great!

PR #632, opened 2023-02-02
Add controlled selection scrolling options

frankagathos:
Added the ability to configure the scrolling when using controlled grid selection.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -468,6 +468,12 @@ export interface DataEditorProps extends Props {
      */
     readonly gridSelection?: GridSelection;
     /**
+     * Scroll position configuration when using {@link gridSelection}.
+     * See "Controlled selection" example for details.
+     * @group Selection
+     */
+    readonly gridSelectionScrollOptions?: GridSelectionScrollOptions;
+    /**
      * Emitted whenever the grid selection changes. Specifying
      * this function will make the grid’s selection controlled, so
      * so you will need to specify {@link gridSelection} as well. See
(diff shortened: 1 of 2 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -467,6 +467,12 @@ export interface DataEditorProps extends Props {
      * @group Selection
      */
     readonly gridSelection?: GridSelection;
+    /**
+     * Scroll position configuration when using {@link gridSelection}.
+     * See "Controlled selection" example for details.
+     * @group Selection
+     */
+    readonly gridSelectionScrollOptions?: GridSelectionScrollOptions;
```
jassmith, line 475: I really like this option, I think we need to workshop the name a bit. Not quite sure I have a better alternative yet... Did you have any other ideas?

frankagathos: maybe a shorter more generic name like: "gridSelectionScroll" works better.

jassmith: `focusScrollBehavior` `focusBehavior` `gridSelectionBehavior`

One of the things about this is it actually only reflects the `gridSelection.current.cell`

frankagathos: I prefer ```gridSelectionBehavior```

jassmith: approved

jassmith: I still need to do an audit of how this interacts with all the other options, would be helpful to break down how this corresponds with the paddingX/Y options etc

frankagathos: @jassmith anything oustanding on my end for this to get merged?

PR #652, opened 2023-02-24
Add onFill to allow overriding the filled value

jimmybergman:
I want to use this like:
```js
  const calculateFill = React.useCallback((fillValue: EditableGridCell, colBefore: any[], targetOffset: number) => {
    if (colBefore.length >= 2 && typeof colBefore[0] === 'number' && typeof colBefore[1] === 'number') {
      // Assume arithmetic series
      const seriesOffset = colBefore[1] - colBefore[0]
      fillValue.data = colBefore[0] + targetOffset * seriesOffset
      return fillValue
    }
  }, [])
  const onFill = React.useCallback(
    (fillValue: ReadWriteGridCell, targetCell: Item, selectedCells?: CellArray) => {
      if (selectedCells && selection?.current?.range) {
        const targetCol = targetCell[0] - selection.current.range.x
        const colBefore = selectedCells.map(r => r[targetCol]?.data)
        return calculateFill(fillValue, colBefore, targetCell[1] - selection.current.range.y)
      }
    },
    [selection, calculateFill]
  )
  ```
  
  to get arithmetic series expanded when filling instead of always getting the value from the top cell.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -15,6 +15,8 @@ import {
     GridKeyEventArgs,
     GridMouseEventArgs,
     GridSelection,
+    CellArray,
+    ReadWriteGridCell,
     isEditableGridCell,
     Rectangle,
     isReadWriteCell,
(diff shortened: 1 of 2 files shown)


jassmith:
I have no fuckin idea why githut wont let me push to this branch but alex can...

Issue #648, opened 2023-02-20
Allow measuring custom cells


LukasMasuch:
Is this line still needed? Theoretically, a custom cell could also just implement the `measure` function to allow being resized correctly (e.g., in the case of `date-picker-cell` this would be very useful):

https://github.com/glideapps/glide-data-grid/blob/c9e1f3df5d545171ccca2226c0c55e253fb6cc83/packages/core/src/data-editor/use-column-sizer.ts#L24

Currently, I think there isn't any way to allow custom cells to be resized to the actual content or?

BrianHung:
I'm implementing a custom "formula" cell type that is similar to TextCell (`customRenderers` seems more DX ergonomic than `drawCell`); this feature would be useful for auto-resizing on double click.

chkn:
Closing since this was fixed in https://github.com/quicktype/glide-data-grid/pull/703

PR #656, opened 2023-03-02
Allow measuring custom cells

LukasMasuch:
In the current state, custom cells cannot be measured and are not auto-sized. This PR removes a check that does not seem to be required anymore:

https://github.com/glideapps/glide-data-grid/blob/c9e1f3df5d545171ccca2226c0c55e253fb6cc83/packages/core/src/data-editor/use-column-sizer.ts#L24

Closes: https://github.com/glideapps/glide-data-grid/issues/648

### DIFF
--- a/packages/cells/src/cell.stories.tsx
+++ b/packages/cells/src/cell.stories.tsx
@@ -399,8 +399,8 @@ export const CustomCells: React.VFC = () => {
                         width: 150,
                     },
                     {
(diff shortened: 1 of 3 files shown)


LukasMasuch:
Resolved by https://github.com/quicktype/glide-data-grid/commit/525f94e1a989b035c57da386e4d15b99cfc39571

PR #663, opened 2023-03-09
feat: custom outside click predicate

vabrahamyanadobe:
There are cases when editors are opened via portals and there is no possibility to add "click-outside-ignore" class to it (e.g. [react-spectrum DatePicker](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html) which renders the calendar at the end of the body, without possibility to add a custom class to it).
This change makes it possible to decide whether it's an outside click or not, and prevent grid to exit the edit mode if it's not the case.


vabrahamyanadobe:
@jassmith FYI

PR #664, opened 2023-03-09
feat: expose the cell location in keyboard event

vabrahamyanadobe:
While implementing custom hotkeys, it's useful to have an access to the cell, on which the keyboard event occurred.
E.g. adding new row when CMD+Enter was pressed on the last row


vabrahamyanadobe:
@jassmith FYI

chkn (approved):
Added a test. Will defer to @jassmith for final approval of the new API

jassmith:
I dont hate this idea. It's cheap to do and doesn't seem to offer any downside. I have fixed the API.

PR #673, opened 2023-03-21
Fix issue with missing ts distribution in cells package

LukasMasuch:
`5.2.2-beta3` and `5.2.2-beta4` do not contain a valid `ts` distribution for the `cells` package. The reason is that the files are within the `src` folder instead of in the `dist/ts` root: 

Screenshot 2023-03-21 at 12 58 25

This is `5.2.2-beta2` which still works fine for me:

Screenshot 2023-03-21 at 12 56 20

This happened after the tests were added to the `cells` packages. I believe we need to exclude the files from the test folder in `tsconfig.types.json`. 

I'm also not sure if the test folder needs to be added to the include here?

https://github.com/glideapps/glide-data-grid/blob/e3b3507c1e00c71ff79cd23a792e43f2933a7b41/packages/cells/tsconfig.json#L3

### DIFF
--- a/packages/cells/tsconfig.types.json
+++ b/packages/cells/tsconfig.types.json
@@ -13,5 +13,12 @@
         "emitDeclarationOnly": true,
         "sourceMap": false
     },
-    "exclude": ["./src/**/*.stories.tsx", "./src/stories/*.tsx", "./src/docs/*.tsx", "./src/setupTests.ts",]
+    "exclude": [
+        "./src/**/*.stories.tsx",
+        "./src/stories/*.tsx",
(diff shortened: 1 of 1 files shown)


chkn:
I just pulled the `5.2.2-beta5` package we published last night and it looks fine to me...
CleanShot 2023-08-15 at 10 27 47@2x

LukasMasuch:
@chkn Just checked with `5.2.2-beta5`. The main glide package is fine, but the mentioned issue is specific to the `cells` package. And this still has this issue. I updated the text above to better reflect that it is specific to the `cells` package.

PR #678, opened 2023-03-27
allow custom editor to choose movement via onFinishedEditing

BrianHung:
Want to support typical spreadsheet editing movements like
- Shift+Enter: Move selection up after closing editor
- Shift+Tab Move selection left after closing editor
- Arrow-up: Move selection up after closing editor if formula is plain text
- Arrow-left, Arrow-down, Arrow-right: same as above for other directions

In general, think it should be left up to developers to decide which direction to go to using the new `onFinishedEditing`, and deprecate the default `onKeyDown` for custom editors.

```
<div className="clip-region" onKeyDown={onKeyDown}>
  {editor}
</div>
```

### DIFF
--- a/packages/core/API.md
+++ b/packages/core/API.md
@@ -506,19 +506,29 @@ onVisibleRegionChanged?: (range: Rectangle) => void;
 ## provideEditor
 
 ```ts
-interface EditorProps {
-    readonly onChange: (newValue: GridCell) => void;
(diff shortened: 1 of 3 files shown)


iamrakhmatov:
@BrianHung does this change enables me to keep focus in the same cell after edit is finished?

PR #693, opened 2023-04-26
feat: pass through middle mouse click event to onCellClicked

hkang1:
# Current State

Currently only the left mouse button event (`button === 0`) is being passed through to the mouse event handler `handleMaybeClick`.

# Proposed Change

Also pass through the middle mouse button through to the cell `onCellClicked` event handler to allow for different behaviors between the two mouse buttons.

cc @jassmith @ivoelbert


hkang1:
any thoughts on this small change to unlock the middle mouse button and passing it through to the hook 🙏

PR #700, opened 2023-05-09
Switch to @faker-js/faker

pzcfg:
It seems that [faker](https://www.npmjs.com/package/faker) is no longer maintained.

[@faker-js/faker](https://www.npmjs.com/package/@faker-js/faker) has the same functionality (and more) as well as built-in typescripting


pzcfg:
placeimg.com has shut down 😢

Screenshot 2023-07-19 at 12 03 52 PM

resulting in broken images in the editor demos

Screenshot 2023-07-19 at 12 04 25 PM

It looks like [@faker-js/faker](https://www.npmjs.com/package/@faker-js/faker) has already switched over to use loremflickr.com instead

Issue #697, opened 2023-05-04
`measureCell` is not allowing custom cells (`GridCellKind.Custom`) to actually calculate the cell width


hkang1:
[Relevant PR](https://github.com/glideapps/glide-data-grid/pull/703)
[Relevant line](https://github.com/glideapps/glide-data-grid/blob/f1d23fe54ea95e92639d4857a51ff57d1555df98/packages/core/src/data-editor/use-column-sizer.ts#L24)

```
function measureCell(
    ctx: CanvasRenderingContext2D,
    cell: GridCell,
    theme: Theme,
    getCellRenderer: GetCellRendererCallback
): number {
    if (cell.kind === GridCellKind.Custom) return defaultSize;

    const r = getCellRenderer(cell);
    return r?.measure?.(ctx, cell, theme) ?? defaultSize;
}
```

The line `if (cell.kind === GridCellKind.Custom) return defaultSize;` here skips any measurement and returns a constant value of `150` when the cell type is of `GridCellKind.Custom`. Shouldn't this line just get removed and allow users to define measuring callbacks?

chkn:
Closing since the fix was merged.

PR #703, opened 2023-05-12
fix: allow measureCell to handle custom cell types

hkang1:
[Issue this PR is addressing](https://github.com/glideapps/glide-data-grid/issues/697)

Problem:
The current `measureCell` function that is used to determine the size of a cell based on various content uses a `defaultValue` of 150px when it encounters a cell type of `GridCellKind.Custom`. This causes all of the custom cell type to arbitrarily auto adjust the custom cell columns to be 150px no matter what.

Proposal:
Remove the custom cell type guard to allow custom cells to define their own `measure` callback to handle how they want to be measured.

### DIFF
--- a/packages/core/src/data-editor/use-column-sizer.ts
+++ b/packages/core/src/data-editor/use-column-sizer.ts
@@ -5,7 +5,6 @@ import type { GetCellRendererCallback } from "../data-grid/cells/cell-types";
 import {
     CellArray,
     GridCell,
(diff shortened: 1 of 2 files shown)


hkang1:
cc @jassmith @ivoelbert

moritzuehling:
We'd love to use this feature, and hope this gets merged (and released, alongside the current main branch) soon!

Issue #722, opened 2023-06-16
Make the demos in documentation more easily runnable / replicable


adamerose:
Following up on issues https://github.com/glideapps/glide-data-grid/issues/538 and https://github.com/glideapps/glide-data-grid/issues/565, I think the documentation could still use some improvements.

These were some pain points for me:

-  Copy pasting code from the storybook sidebar always includes the line numbers
- The storybook code shows relative file imports which have to be replaced, and the hierarchy isn't consistent with the NPM distribution so for example replacing the first snippet with the second one here is not obvious.
```
import { DataEditor } from "../../data-editor/data-editor"
import { GridCellKind } from "../../data-grid/data-grid-types"
```
```
import DataEditor, { GridCellKind } from "@glideapps/glide-data-grid
````
- A lot of examples use utilities like `useMockDataGenerator` and `defaultProps` that aren't in the distribution and aren't trivial to replace. I think most of the demos don't really need a data generator utility and could just explicitly define a data object and related functions, like in this [Editing Data](https://glideapps.github.io/glide-data-grid/?path=/story/glide-data-grid-docs--editing-data) demo, which is much easier to understand and replicate.

pzcfg:
> Copy pasting code from the storybook sidebar always includes the line numbers

The code display/rendering comes from the `storysource` add-on in https://github.com/storybookjs/storybook

It looks like this was actually fixed in v7.0.0 https://github.com/storybookjs/storybook/pull/18725 https://github.com/storybookjs/storybook/issues/18629

I took a stab #724 at [upgrading storybook](https://storybook.js.org/docs/react/configure/upgrading#upgrade-script) using their `npx storybook@latest upgrade` and the handful of demos I tried worked and the code copied without line numbers, but it could definitely use more testing.

> The storybook code shows relative file imports which have to be replaced, and the hierarchy isn't consistent with the NPM distribution so for example replacing the first snippet with the second one here is not obvious.

Not sure what the best solution is here. Possibly breaking the examples out into their own package, similar to how `cells`, `core`, and `source` are, but not sure what the implications are there for development and testing.

> A lot of examples use utilities like useMockDataGenerator and defaultProps that aren't in the distribution and aren't trivial to replace. I think most of the demos don't really need a data generator utility and could just explicitly define a data object and related functions, like in this [Editing Data](https://glideapps.github.io/glide-data-grid/?path=/story/glide-data-grid-docs--editing-data) demo, which is much easier to understand and replicate.

[...]

PR #724, opened 2023-06-16
Upgrade storybook to v7.0.21

pzcfg:
This updates storybook to v7.0.21 using [their script](https://storybook.js.org/docs/react/configure/upgrading#upgrade-script) `npx storybook@latest upgrade` and [required manual migration](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#removed-global-client-apis).

This is primarily to fix the line numbering issue mentioned in #722

I tested a handful of demos, which worked, and the code copied without line numbers, but it could definitely use more testing.


jassmith:
Upgrade is complete

PR #736, opened 2023-07-10
feat: allow custom cells to define their own measuring methods

hkang1:
Current glide table does not allow for `measureCell` functions to run properly for custom cells (only default cells support this). This allows for custom cells to work as well.


hkang1:
duplicate

PR #743, opened 2023-07-19
Add `id` attribute to cell editor input

### LINKED ISSUE #742: A form field element should have an `id` or `name` attribute
The following warning appears in the browser whenever a cell is edited:

> A form field element should have an `id` or `name` attribute

package version: `5.2.1`
browser: Chrome 114

https://github.com/glideapps/glide-data-grid/assets/25715951/bd742b7f-dae9-4765-b419-19e7df2718c7

romanstetsyk:
Fixes #742 
Please review this PR

### DIFF
--- a/packages/core/src/growing-entry/growing-entry.tsx
+++ b/packages/core/src/growing-entry/growing-entry.tsx
@@ -12,6 +12,8 @@ interface Props
     readonly validatedSelection?: SelectionRange;
 }
 
+let globalInputID = 0;
+
 /** @category Renderers */
 export const GrowingEntry: React.FunctionComponent<Props> = (props: Props) => {
     const { placeholder, value, onKeyDown, highlight, altNewline, validatedSelection, ...rest } = props;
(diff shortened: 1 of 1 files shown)


packages/core/src/growing-entry/growing-entry.tsx
```diff
@@ -23,6 +23,8 @@ export const GrowingEntry: React.FunctionComponent<Props> = (props: Props) => {
 
     assert(onChange !== undefined, "GrowingEntry must be a controlled input area");
 
+    const [inputID] = React.useState(() => "input-box-" + Math.round(Math.random() * 1000));
```
pzcfg, line 26: This should probably use something like `useId()` instead https://react.dev/reference/react/useId

romanstetsyk: That was my first thought to use `useId()`, but I decided to make it similar to [`searchID`](https://github.com/quicktype/glide-data-grid/blob/c9e1f3df5d545171ccca2226c0c55e253fb6cc83/packages/core/src/data-grid-search/data-grid-search.tsx#L78).
Would you like me to update it with `useId`?

pzcfg: Oh hrm, I feel like that should be using `useId()` as well, but honestly not really sure if there's a good reason to / not to. In general I feel like it's better to avoid peppering `Math.round(Math.random() * 1000)`-type things around though.

romanstetsyk: Oh, it appears that there's no `useId` in `"react": "^16.12.0 || 17.x"` so we can't use it, unfortunately.

PR #755, opened 2023-08-15
fix: disable autoscrolling when dragging column headers

julian-determined-ai:
[Issue](https://github.com/quicktype/glide-data-grid/issues/754)

Problem:
Currently, pressing down and holding a column header will cause the table to unexpectedly vertically scroll scroll.  

An example can be seen below: 

https://github.com/quicktype/glide-data-grid/assets/103522725/7656a4db-a9ae-48a0-9e52-f7f835eebdda

Proposal:
Update the `onMouseMove` logic to so that the table does not scroll vertically when a column header is pressed down. Horizontal scrolling within the table should still be enabled so that users can properly drag and drop column headers.

This PR accomplishes the above by storing and checking when a `header` is being pressed down, if a `header` is being pressed, then we only track the horizontal scroll direction but do not set a vertical scroll direction.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -1818,7 +1818,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
             themeForCell,
         ]
     );
-
+    const isActivelyDraggingHeader = React.useRef(false);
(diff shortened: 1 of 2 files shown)


julian-determined-ai:
cc: @jassmith @ivoelbert

jassmith (approved):

PR #758, opened 2023-08-18
5.3.0

jassmith:
## Bug Fixes

- Compatibility fixes with next-js
- Boolean cells respect padding theme options better
- Uri cells respect padding theme options better
- Fix issue where markdown editor would not report the edit completion event correctly
- Color parsing no longer crashes when receiving unknown colors.
- Resize end no longer triggers spuriously
- Fix some cases where out of bounds cells may be requested
- Bubble and dropdown cells handle commas and empty arrays better
- Fix some crashers
- Image cell now respects content alignment
- Fix some cases where `onDelete` would not be triggered.
- Fix a bug where column resizing would be spuriously allowed on row marker column edge. Thank you @BrianHung
- Fix a bug where selection could end up out of range if the number of columns or rows decreased
- Fix column resizing when canvas is scaled. Thank you @BrianHung
- Fix scrollTo when canvas is scaled. Thank you @BrianHung
- Fix a bug where the last column cannot always be selected
- Click off behavior now works better with touch support
- Fix some weird behaviors with rtl mode browsers
- Fix bug where final column could not always be copied
- Column headers now support RTL mode better
- Checkbox now property respects content alignment
- Wrapped text now better suppor

## New Features

- New hook: `useUndoRedo` adds support for in memory undo and redo
- Row markers now can be configured with `checkbox-visible` for always visible row markers
- Boolean cells now support a max size to prevent overly large checkboxes with large rows
- Uri cells now support the display data property which can be used to format the url displayed to the user
- Number cells now have the ability to specify a `fixedDecimals` for a fixed number of display decimals
- Number cells now have the ability to specify if negative numbers are allowed using the `allowNegative` property
- `copyData` can now be set on any cell. This overrides the natural copy behavior.
- DataEditor ref now contains a callback to `remeasureColumns`. This allows implementors to force the data grid to recompute column sizes whenever they want.
- Search functionality can now be overriden by implementors using the `searchResults`, `searchValue`, and `onSearchValueChange` properties.


schickling:
HE IS BACK!

PR #787, opened 2023-10-04
Fix support for `displayData` in uri cell

LukasMasuch:
The URI cell allows configuring display data, but the behavior differs from other cell types. The display data is only applied to the overlay editor, but not drawn to the canvas. This PR fixes this by preferring `displayData` in case it is configured.

### DIFF
--- a/packages/core/src/data-grid/cells/uri-cell.tsx
+++ b/packages/core/src/data-grid/cells/uri-cell.tsx
@@ -13,7 +13,7 @@ export const uriCellRenderer: InternalCellRenderer<UriCell> = {
     useLabel: true,
     drawPrep: prepTextCell,
     draw: a => drawTextCell(a, a.cell.displayData ?? a.cell.data, a.cell.contentAlign),
-    measure: (ctx, cell, theme) => ctx.measureText(a.cell.displayData ?? cell.data).width + theme.cellHorizontalPadding * 2,
+    measure: (ctx, cell, theme) => ctx.measureText(cell.displayData ?? cell.data).width + theme.cellHorizontalPadding * 2,
     onDelete: c => ({
         ...c,
         data: "",
--- a/packages/core/src/data-grid/cells/uri-cell.tsx
+++ b/packages/core/src/data-grid/cells/uri-cell.tsx
@@ -12,8 +12,8 @@ export const uriCellRenderer: InternalCellRenderer<UriCell> = {
     needsHoverPosition: false,
     useLabel: true,
     drawPrep: prepTextCell,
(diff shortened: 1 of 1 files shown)


jassmith:
Im a bit wishy/washy on this one, to fix this correctly I think we need to make this draw as a url (underline, blue font?) if we are going to preview with the display data. It also leads to an impression that it should be single clickable.

jassmith:
That raises a new even more fun issue, canvas doesn't really properly support underline text

LukasMasuch:
@jassmith Oh indeed, I see the issue 😬 I actually have that on my list to investigate if we can improve the URI cell by adding an underline and using the configured link color in rendering. I have seen a version of this [here](https://github.com/determined-ai/determined/blob/main/webui/react/src/pages/F_ExpList/glide-table/custom-renderers/cells/linkCell.tsx), which attempts to add underline but makes it configurable via an offset. Maybe there is even a way to automatically determine a good underline offset based on the font + size, but I'm not sure ... 

The [implementation](https://github.com/determined-ai/determined/blob/main/webui/react/src/pages/F_ExpList/glide-table/custom-renderers/cells/linkCell.tsx) also adds the capability to click on the link directly, but I think this could be something to provide as an optional feature.

LukasMasuch:
Hmm, since the implementation is using `getMiddleCenterBias`, it might already adapt well to different fonts and sizes.

jassmith:
I already implemented the underline in the link cell in the cells package. The problem is its a bit slow to make it work well. The line needs to break for descenders, which I implemented, but still, slow.

LukasMasuch:
Oh nice, I didn't know about link cell 👍 Is the descenders break what makes it slow? Does it need to have the descenders break? I tried underlining in a couple of apps, and some just display it under the text. For example, this is in Notion:

image

LukasMasuch:
> It's definitely significantly better but its hard to make reliably with arbitrary fonts

Maybe we could determine the descenders line by doing some calculation in a similar way as done in `getMiddleCenterBias` by using a sample with a descenders characters? E.g., if we compare a sample with and one without a descender character we could figure out the relative difference based on the font.

jassmith:
This can kind of be made to work but its not as reliable as that method makes it seem. We can try. Honestly I am okay with the perf hit if we code this very very carefully. This is one of those cases where we want to make a test page where we can microbench it and then just try 10 different things and see what is actually fastest. It's going to be one of those cases where one solution will end up being 10x faster.

jassmith:
I've taken this PR as is into 6.0.0, sorry for not merging, I couldn't get github to let me commit the merge commit to your branch so this would merge cleanly :(

Issue #784, opened 2023-10-02
Return total scroll width & height via getBounds


LukasMasuch:
Is there any way to figure out the current height and width of the scroll area? I need to determine if the table currently requires scrolling.  

My suggestion would be to adapt the `getBounds` on the dataGridRef to return the full height & width of the scroll area (`dvn-stack`) if no `col` and `row` is specified.

LukasMasuch:
I did a draft implementation [here](https://github.com/glideapps/glide-data-grid/pull/788) on how this could look like.

LukasMasuch:
Solved by https://github.com/glideapps/glide-data-grid/pull/788

PR #788, opened 2023-10-05
Add support for returning bounds of entire scroll area

LukasMasuch:
This is a rough draft of how the entire scroll bounds could be returned via `getBounds` if `cell` and `row` are undefined (see https://github.com/glideapps/glide-data-grid/issues/784). I did some manual testing, and it seems to align well with what is returned by `getBounds` for specific items. The height differs by one pixel when compared with the last cell bounds. But that is probably the border that is also included in the scroll area.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -3443,11 +3443,12 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 if (col === undefined && row === undefined && scrollRef.current && canvasRef.current) {
                     // Return the bounds of the entire scroll area:
                     const rect = canvasRef.current.getBoundingClientRect()
+                    const scale = rect.width / scrollRef.current.clientWidth
                     return {   
-                         x: rect.x - scrollRef.current.scrollLeft,
-                         y: rect.y - scrollRef.current.scrollTop,
-                         width: scrollRef.current.scrollWidth,
-                         height: scrollRef.current.scrollHeight,
+                         x: rect.x - scrollRef.current.scrollLeft * scale,
+                         y: rect.y - scrollRef.current.scrollTop * scale,
+                         width: scrollRef.current.scrollWidth * scale,
+                         height: scrollRef.current.scrollHeight * scale,
                      };
                 }
                 return gridRef.current?.getBounds(col !== undefined ? col + rowMarkerOffset : undefined, row);
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -3440,7 +3440,17 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 return gridRef.current?.damage(damageList);
             },
             getBounds: (col, row) => {
-                return gridRef.current?.getBounds(col + rowMarkerOffset, row);
+                if (col === undefined && row === undefined && scrollRef.current && canvasRef.current) {
+                    // Return the bounds of the entire scroll area:
+                    const rect = canvasRef.current.getBoundingClientRect()
+                    return {   
+                         x: rect.x - scrollRef.current.scrollLeft,
+                         y: rect.y - scrollRef.current.scrollTop,
+                         width: scrollRef.current.scrollWidth,
+                         height: scrollRef.current.scrollHeight,
+                     };
+                }
+                return gridRef.current?.getBounds(col !== undefined ? col + rowMarkerOffset : undefined, row);
             },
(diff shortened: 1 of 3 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3440,7 +3440,18 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                    const rect = canvasRef.current.getBoundingClientRect()
+                    const scale = rect.width / scrollRef.current.clientWidth
+                    return {   
+                         x: rect.x - scrollRef.current.scrollLeft * scale,
+                         y: rect.y - scrollRef.current.scrollTop * scale,
+                         width: scrollRef.current.scrollWidth * scale,
+                         height: scrollRef.current.scrollHeight * scale,
+                     };
+                }
+                return gridRef.current?.getBounds(col !== undefined ? col + rowMarkerOffset : undefined, row);
```
jassmith, line 3454: It would probably be clearer just to keep the 0 here rather than push down the default to the dataGridRef

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3440,7 +3440,18 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 return gridRef.current?.damage(damageList);
             },
             getBounds: (col, row) => {
-                return gridRef.current?.getBounds(col + rowMarkerOffset, row);
+                if (col === undefined && row === undefined && scrollRef.current && canvasRef.current) {
+                    // Return the bounds of the entire scroll area:
+                    const rect = canvasRef.current.getBoundingClientRect()
```
jassmith, line 3445: Maybe this path should return a default value if canvasRef.current is undefined instead of getting the bounds of the first column?

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3440,7 +3440,18 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                 return gridRef.current?.damage(damageList);
             },
             getBounds: (col, row) => {
-                return gridRef.current?.getBounds(col + rowMarkerOffset, row);
+                if (col === undefined && row === undefined && scrollRef.current && canvasRef.current) {
```
jassmith, line 3443: This wont pass linting, explicit booleans are required for checks, in this case `canvasRef.current !== undefined`

jassmith (changes requested):
Looks good, needs a couple tweaks.

packages/core/src/data-editor/data-editor.tsx
LukasMasuch: Changed it to return `undefined` in case canvasRef.current or scrollRef.current is undefined.

packages/core/src/data-editor/data-editor.tsx
LukasMasuch: Fixed it 👍

packages/core/src/data-editor/data-editor.tsx
LukasMasuch: ok, changed it to use `col ?? 0 + rowMarkerOffset` here 👍

packages/core/src/data-grid/data-grid.tsx
```diff
@@ -1499,12 +1499,12 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
...
                 }
             },
-            getBounds: (col: number, row?: number) => {
+            getBounds: (col?: number, row?: number) => {
                 if (canvasRef === undefined || canvasRef.current === null) {
                     return undefined;
                 }
 
-                return getBoundsForItem(canvasRef.current, col, row ?? -1);
+                return getBoundsForItem(canvasRef.current, col ?? 0, row ?? -1);
```
jassmith, line 1507: isnt this duplicate now?

packages/core/src/data-grid/data-grid.tsx
```diff
@@ -283,7 +283,7 @@ type DamageUpdateList = readonly {
 
 export interface DataGridRef {
     focus: () => void;
-    getBounds: (col: number, row?: number) => Rectangle | undefined;
+    getBounds: (col?: number, row?: number) => Rectangle | undefined;
```
jassmith, line 286: oh i see this is shared... hmmmmm

LukasMasuch: yep, the interface is shared 👍  So both need to somehow handle `col===undefined`. We could also return `undefined` in `getBoundsForItem` if `col` is `undefined`.

PR #804, opened 2023-11-08
Convert `&nbsp;` to regular spaces in `decodeHTML`

chkn:
Fixes https://github.com/glideapps/glide/issues/23589

When pasting `text/html` content, we generate a DOM fragment and then `decodeHTML` walks that DOM, reading out the text values of cells. During this process, regular space characters are converted into non-breaking space Unicode characters, which can cause subtle and annoying issues if, for instance the pasted text is re-copied out of GDG.

This PR modifies `decodeHTML` to normalize the strings it reads from the DOM by replacing the non-breaking space characters with regular spaces.

### DIFF
--- a/packages/core/src/data-editor/data-editor-fns.ts
+++ b/packages/core/src/data-editor/data-editor-fns.ts
@@ -173,6 +173,11 @@ export function unquote(str: string): string[][] {
     return result;
 }
 
+function normalizeHTMLString(input: string): string {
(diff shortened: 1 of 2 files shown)


marcel-at-glide:
Curious to know when this fix will go out.

rwdaigle:
@jassmith @chkn any updates here?

jassmith:
This should be fixed already in 5.3.1

PR #810, opened 2023-11-26
6.0.0

jassmith:
> 🚧 These release notes are under construction. They may contain misinformation, partial truths, and damned lies.

### New dependency
 
- canvas-hypertxt bumped to 1.0.3
- @linaria/react now added as dep (extremely tiny)

### ✌️ Farewell create react app 4

CRA 4 is no longer officially supported. While it is definitely possible to make it work, it may require extra work. CRA 5 works fine.

### 🚢 Better exports

Exports for Glide Data Grid are now done in a more standard esm compliant manner. This should enable better tree shaking.

### 🧼 Cell API cleanup

Some cells have had minor tweaks to their API to bring them in line with standard conventions. The `ImageCell` and all the cells in the `cells` package now uses the standard `readonly` flag.

### 👋 Minimap removed

The minimap was awesome, and largely unused. It has been removed from this version of Glide Data Grid in the service of smaller packages and a more maintainable surface area. We appreciate your service 🫡

### 🎨 `drawCell` callback improved.

The drawCell callback now receives a callback argument which paints the standard cell. This allows dramatically more flexibility with custom under and overdrawing of cells. It is also now a void method eliminating a source of confusion.

```typescript
const drawCell: DrawCellCallback = React.useCallback(
    (args, draw) => {
      // Draw something below the cell drawing
      draw()
      // Draw something on top of the cell drawing
}, [])
```

### 💪 Kinetic super scroll on iOS

Are you the one person presenting data grids to your mobile users? This feature is for you. GDG will now artificially boost the framerate during kinetic scroll on iOS to maintain a smooth experience. This is still experimental, but if feedback goes well we intend to promote this to stable.

```tsx
    return (
        <DataEditor
            {...otherProps}
            experimental={{
                kineticScrollPerfHack: true,
            }}
        />
    );
```

### 🏁 Performance improvements when updating lots of data at once

Damage rendering now is 2x faster than the 5.0 series. This is enabled by reducing the amount of clipping during a damage pass. This improvement is what has made the DOOM easter egg on the main site possible. The easter egg may also be the reason for the improvement...

CleanShot 2024-01-10 at 22 43 12@2x

### 🌐 UriCell improvements

CleanShot 2024-01-10 at 22 41 10@2x

The URI cell supports drawing an underscore on hovering a URI value via the `hoverEffect` property. It also supports click events (e.g., to open the URL) via the `onUriClick` property and uses the `displayData` within the cell rendering. 

### ⚽️ Add support for rounding radius in the theme

image

Configure the rounding radii of checkboxes, bubbles, skeletons, images, buttons, and other rounded elements via the optional `roundingRadius` theming property. 

### 🤕 Header menu icons can now be configured

image

Customize the header menu icon via the `menuIcon` property in `GridColumn`. It supports `triangle` for the default menu icon, `dots` for the three-dots menu icon, or a key of one of the icons passed to the `headerIcons` prop.

### 📈 Sparkline cell now supports even more charts

CleanShot 2024-01-10 at 22 37 25@2x

The `line` graphKind got renamed to `area`. The new `line` chart now renders just a simple line chart. Additionally, the x-axis can be hidden via `hideAxis`.

### 🥶 Freeze trailing rows

CleanShot 2024-01-10 at 22 36 58@2x

Get going by specifying the number of freeze trailing rows you want.

```tsx
    return (
        <DataEditor
            {...otherProps}
            freezeTrailingRows={2}
        />
    );
```

### 💀 LoadingCell skeletons

CleanShot 2024-01-10 at 22 34 25@2x

The loading cell can be configured to show skeletons when `skeletonWidth` is set to a value > 0. The height can be configured via the optional `skeletonHeight` property, and it also supports a randomized variability applied to the width via `skeletonWidthVariability.` 

### ⌨️ Fully remappable keybindings

Keybindings can now be remapped instead of just turned off and on. More details [here](https://docs.grid.glideapps.com/api/dataeditor/input-interaction#keybindings).

### 🪤 Focus trapping

There is now a `trapFocus` prop that will cause the grid to prevent focus leaving the grid during caret browsing or pressing tab.

### 🙅‍♀️ Prevent column reordering

Reordering of columns can now be prevented as new column locations are proposed using the new `onColumnProposeMove` callback. This allows for greater control over where users are allowed to drag columns.

### 💾 Copy and paste methods now available

The copy and paste functionality of the data grid is now exposed for direct usage by developers.

### 🌲 Tree view cell 

image

This cell represents the basic building block required to make collapsable and groupable rows. We will continue to improve support for row grouping in the rest of the 6.0.0 series.

### ⬇️ Dropdown cell improvements

Allow specifying label and value independently in dropdown cell and some other visual improvements.

### 💍 Improved drawing of highlight and selection rings

When the selection ring or a highlight ring is at the right or bottom edge of the grid it will no longer be clipped out.

### 🦏 Safari performance improvements

Safari now renders around 2 to 5 times faster on mobile devices. Less hitching, better grids, papa johns.

### 🔦 Large highlight region support

Prior to 6.0.0 large highlight regions could cause excessively slow drawing performance. This is no longer a problem.

### 🔍 Search no longer renders eagerly

Don't use the search? No problem. You no longer pay the penalty for something you don't need.

### 🏁 Mouse hover render reduction

Hovering with the mouse no longer results in excessive react rendering.

### 👓 Hover interactions now can discern blank spots of grid from outside of grid

More details here: https://github.com/glideapps/glide-data-grid/pull/811

### ↔️ Resize column indicator

image

There is a new column resize indicator. It can be configured via the `resizeIndicatorColor` theme property. 

### ☕️ Fill handle improvements

- Controllable selection behavior
- FillPattern support
- Larger fill handle click region
- New visual representation

CleanShot 2024-01-10 at 23 10 57@2x

### 🔲 Get the bounds of the entire scroll area 

If `getBounds` get called with `col` and `row` as undefined, the bounding box of the entire data grid scroll area is returned.

## 🔚 Odds and ends

- GDG now prefixes most of the css classes it uses to avoid conflicting with commonly used names.
- Page up and down keybindings are now on by default

## 🐞 Bug Fixes

- `onDelete` now properly called when doing a cut operation.
- Fixed a crasher caused by attemping to draw negative radius arcs.
- Fix a bug where `getBounds` would compute for the wrong cell.
- Fixed multiple bugs where strict mode would be violated
- Horizontally center bubble cell editor to align with cell rendering.
- Fix copy escaping for array values. 
- Fix issues with paste logic.
- Fix text cell editor color in Safari.
- Fix jittering issue with grow columns.
- Use the configured padding for measuring of number, row-id, dropdown and date-picker cells.
- Fix sparkline chart error related to arrays with less than 3 values. 

```[tasklist]
### Merge blockers
- [x] The focus ring does not draw over the sticky row
- [x] Highlight rings do not draw on the header
- [x] Hovering sticky row destroyed focus ring
- [x] Merge #831
- [x] Merge #830
- [x] Freeze rows jitter in ios due to double buffering
```


LukasMasuch:
@jassmith Thanks for this update 🙏  I quickly checked the latest alpha, and almost everything seems to work fine. The only aspect that seems to be broken is the column header icon: 

With 5.99.9-alpha5: 

Screenshot 2023-11-29 at 16 30 47

With 5.3.1:

Screenshot 2023-11-29 at 16 33 26

jassmith:
Do the default header icons work for you?

LukasMasuch:
Yep, the default header icons work fine. Just custom ones seem to be broken in my setup.

jassmith:
Found the issue

LukasMasuch:
I can confirm that custom icons in `5.99.9-alpha6` are working again 👍

LukasMasuch:
I tested the latest beta (`5.99.9-beta3`), and everything seems to work fine for our setup 👍 Thanks for adding back `drawCells` and updating the UriCell. After some testing, I have a bit of feedback related to UriCell version in this branch:

1. With the added support for directly clicking on links, our preferred behavior for editing would be if the URI overlay editor directly jumps into editing mode instead of the current two step process. Maybe it would be possible to add a cell parameter that would force the editing mode. There are a couple of reasons why we prefer that behavior 1) its more aligned with the behavior of other cells 2) The edit pen icon cannot be changed, and it doesn't fit our icon set 3) there seems to be issues with the UriCell overlay if it is outside of the table boundaries (e.g. the edit icon cannot be clicked). Here is our custom cell setup:

https://github.com/glideapps/glide-data-grid/assets/2852129/83537eff-8b6d-448a-a900-af7c90af6013

2. The text in the cell and within the overlay aren't completely aligned when you open the overlay. A quick fix here is by just adding a `padding-bottom: 3px;` to `gdg-link-area` (that's what we did, but maybe there is also a less hacky fix). 

3. The hover underline works fine for us. But from my tests with different fonts, I think the version in `links-cells` using `getMiddleCenterBias + offset of 5` seems to be even closer to the HTML version. 

4. In general, it would be quite useful if all cell types would officially support an empty state (`data` is `null`/`undefined`). At least for our use cases, there is a semantical difference when a cell has an empty string as vaue or if the value is actually missing. It seems like most cell types are implemented in a way that it actually supports setting it to `null`, but it is not really something officially supported as part of the types.

jassmith:
1) Yeah we want to get rid of the dual mode of that entirely.

2) Hopefully when we fix the overlay this issue goes away too.

3) I haven't worked out why yet but + 5 looks way too much in URI cell but fine in Links Cell. I need to work it out. I will probably make this a theme value so it can be tweaked based on need.

4) I don't understand why you need this information to be known to the data grid itself?

jassmith:
I figured out what you were trying to tell me, sorry wasn't getting through. Fixed.

LukasMasuch:
I did some tests with the latest beta (`5.99.9-beta8`). Overall, almost everything works fine. 

I really like the change that the borders of the last cells are drawn differently; it cleans up some hacky stuff on our side. But there seems to be an issue with **readonly** custom cells and checkbox cells (maybe related to the hover?):

https://github.com/glideapps/glide-data-grid/assets/2852129/027f7d20-9357-4030-a368-4c0090a211c7

It also doesn't work with span cells, but that is probably a very niche case. 

Another slightly unexpected behavior is the usage of `rangeSelect="none"`. Previously, we used this value to disable range selection in some scenarios, but with the latest beta it seems to allow range selection via hotkeys. However, `"cell"` is activating the behavior we want, so we just switched to this.

LukasMasuch:
Another potential issue we just encountered during browser window resizing is `Maximum update depth exceeded` from the `InfiniteScroller`:

image

The stack related to this issue: 

https://github.com/glideapps/glide-data-grid/blob/2f35e56d7939b41dccc0878aa89252a0809c7c9f/packages/core/src/common/utils.tsx#L242

https://github.com/glideapps/glide-data-grid/blob/2f35e56d7939b41dccc0878aa89252a0809c7c9f/packages/core/src/data-editor/data-editor.tsx#L2392

https://github.com/glideapps/glide-data-grid/blob/2f35e56d7939b41dccc0878aa89252a0809c7c9f/packages/core/src/internal/scrolling-data-grid/scrolling-data-grid.tsx#L221

https://github.com/glideapps/glide-data-grid/blob/2f35e56d7939b41dccc0878aa89252a0809c7c9f/packages/core/src/internal/scrolling-data-grid/scrolling-data-grid.tsx#L244

https://github.com/glideapps/glide-data-grid/blob/8e73eb57a1135645a3989255fabbe182cc2bcf0e/packages/core/src/internal/scrolling-data-grid/infinite-scroller.tsx#L215

https://github.com/glideapps/glide-data-grid/blob/8e73eb57a1135645a3989255fabbe182cc2bcf0e/packages/core/src/internal/scrolling-data-grid/infinite-scroller.tsx#L233

jassmith:
I am unable to reproduce the max update depth issue. Are you maybe recreating an input arg on render when you shouldn't? Not saying we shouldn't be permissive of it, I just need to figure out the repro case.

LukasMasuch:
**Update on the issue above:** It seems to be caused on our side if the table width is set to a subpixel width (e.g. 300.5). Not exactly sure why this is happening, but rounding the width solves it for us.

jassmith:
reported issues should be resolved.

LukasMasuch:
@jassmith I did a quick test with the `5.99.9-charlie2`. The `GridMouseEventArgs` doesn't seem to be exported anymore. Is there an alternative to import this type? 

And something seems to be broken with the highlight ring + cell selection on frozen columns:

https://github.com/glideapps/glide-data-grid/assets/2852129/9f1df47b-dc8a-4fbc-8ebf-13a60d562555

Stacktrace:

```
log.ts:48 TypeError: Cannot read properties of undefined (reading 'width')
TypeError: Cannot read properties of undefined (reading 'width')
    at computeBounds (data-grid-lib.ts:725:1)
    at data-grid-render.tsx:1689:1
    at Array.map (<anonymous>)
    at drawHighlightRings (data-grid-render.tsx:1653:1)
    at drawGrid (data-grid-render.tsx:2568:1)
    at data-grid.tsx:770:1
    at data-grid.tsx:826:1
    at commitHookEffectListMount (react-dom.development.js:20573:1)
    at commitLifeCycles (react-dom.development.js:20634:1)
    at commitLayoutEffects (react-dom.development.js:23426:1)
```

jassmith:
Export fixed

jassmith:
The bugs you found lucas are edge cases of the scrolling improvements. PR in 2 mins

jassmith:
Ok so as a last point of effort here we need to go through the changes and make sure we have a comprehensive set of release notes.

jassmith:
I've added a very very very rough draft to the PR description. This needs to be cleaned up and have images added, code samples where needed, etc.

LukasMasuch:
@jassmith Just tried out `5.99.9-charlie3` and still sometimes getting this error during cell selection (not anymore related to a frozen column):

https://github.com/glideapps/glide-data-grid/assets/2852129/f034bfde-b65a-4989-add8-885b7a26ea33

```
Uncaught TypeError: Cannot read properties of undefined (reading 'width')
    at computeBounds (data-grid-lib.ts:722:1)
    at data-grid-render.tsx:1691:1
    at Array.map (<anonymous>)
    at drawHighlightRings (data-grid-render.tsx:1655:1)
    at drawGrid (data-grid-render.tsx:2558:1)
    at data-grid.tsx:769:1
    at data-grid.tsx:825:1
    at commitHookEffectListMount (react-dom.development.js:20573:1)
    at commitLifeCycles (react-dom.development.js:20634:1)
    at commitLayoutEffects (react-dom.development.js:23426:1)
```

LukasMasuch:
The new border around the highlighted area is also sometimes cut off. In the previous version, it was on all sides and now it seems to only been cut off at the bottom:

https://github.com/glideapps/glide-data-grid/assets/2852129/6fddeab5-2e39-4f81-8c38-a90e5cfb40c6

And the additional border for frozen columns is expected, or?

https://github.com/glideapps/glide-data-grid/assets/2852129/c35ef31c-bb24-4e37-be5e-c742135294fc

jassmith:
The extra border is expeted, the cut off is not. Looking into it.

jassmith:
Can you validate the charlie5 release fixes your crasher?

LukasMasuch:
It feels a bit harder, but it's still crashing :( 

https://github.com/glideapps/glide-data-grid/assets/2852129/99a3502c-cc6d-492d-a178-0c64db5faca2

jassmith:
Do you have the latest trace by chance?

LukasMasuch:
Yep:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'width')
    at computeBounds (data-grid-lib.ts:743:1)
    at data-grid-render.tsx:1670:1
    at Array.map (<anonymous>)
    at data-grid-render.tsx:1668:1
    at Array.map (<anonymous>)
    at drawHighlightRings (data-grid-render.tsx:1664:1)
    at drawGrid (data-grid-render.tsx:2492:1)
    at data-grid.tsx:771:1
    at data-grid.tsx:828:1
    at commitHookEffectListMount (react-dom.development.js:20573:1)
```

jassmith:
Can you tell me what col, row it is trying to compute the bounds of maybe?

LukasMasuch:
Before the error occurs, it is trying to call `computeBounds` with col 6 and row 1.

LukasMasuch:
I can also reproduce this with this storybook story in cells:

https://github.com/glideapps/glide-data-grid/assets/2852129/caa0c493-d60c-4862-9afc-2e934e0a2cdc

jassmith:
Wtf why can I not repro this no matter what I try...

jassmith:
Ok I finally got it repro'd. You have to mouse over the right edge. This is fixed already in 6.0.0, will put out charlie6 for confirmation.

LukasMasuch:
woohoo :) I can't get it to break with `v5.99.9-charlie7`

But there is another strange thing going on with the drag and drop :(

https://github.com/glideapps/glide-data-grid/assets/2852129/6d573e88-530e-4232-83cb-25c2ae20a16d

It somehow doesn't clear up the first selection in the second table... and always jumps back to it or is also reacting when I'm actually in another table.

jassmith:
oh I think I know what causes this. Easy enough fix.

jassmith:
Ok more on that. It is just caused by the mouse going off the window and then releasing the drag. The data editor then cannot see the mouse up event and ends up in a bad state.

jassmith:
I think we should move into release mode. It is time to clean up the notes, fix the docs, and ship this puppy.

LukasMasuch:
@jassmith I did another round of testing, and only came across one aspect that is a bit unexpected:

https://github.com/glideapps/glide-data-grid/assets/2852129/710ab954-d3a4-4e57-88ba-3a95bd5e60be

These borders when updating or changing cells are probably not expected? But not sure why this happens, I'm not able to reproduce this in the glide-data-grid storyboard

LukasMasuch (approved):
Everything else looks great 👍

jassmith:
It is reproducable on the Small Editable Grid story

jassmith:
Fixed

PR #811, opened 2023-11-27
fix: manage direction of hover when grid table smaller than grid

luciacangarova:
Hi! I have a use-case where the canvas is stretched over the whole screen but sometimes there are not enough rows/columns to fully fill the grid in all directions. The current behavior of the grid is to create an empty space which is classified as an `out-of-bound` cell with direction `[0,0]`. 
To be able to exactly pinpoint right / bottom / bottom-right, I would like to know the inner direction too based on the col and row.
I suggest doing it by adding a new field, `innerDirection`, to the `GridMouseOutOfBoundsEventArgs` payload, which in conjunction with the `direction` field, allows us to clearly identify the source of the mouse event.

In addition, I also would like to trigger the `onItemHover` function whenever the direction or the inner direction of my "out-of-bound" cell changes - which I also included as part of this PR.

Please let me know if you agree with this approach and the new names.

Thank you!


jassmith:
Can you give me a quick sample code of how you want to use this?

packages/core/src/data-grid/data-grid-types.ts
```diff
@@ -127,6 +127,7 @@ export interface GridMouseOutOfBoundsEventArgs extends BaseGridMouseEventArgs {
     readonly location: Item;
     readonly direction: readonly [-1 | 0 | 1, -1 | 0 | 1];
     readonly isMaybeScrollbar: boolean;
+    readonly innerDirection: readonly [-1 | 0 | 1, -1 | 0 | 1];
```
jassmith, line 130: I've been thinking about this and I just don't like the name. Do you have any other suggestions to help us find a better name?

luciacangarova:
> I think this makes a lot more sense, what do you think?

That is awesome! Thank you 😊 It makes more sense now

PR #822, opened 2023-12-12
Fix type signature of onVisibleRegionChanged prop

Macil:
The `onVisibleRegionChanged` prop has a type signature saying that the `tx`, `ty`, and `extras` parameters are optional, in effect claiming that glide grid will not always pass them to the user's callback, but they are in fact always passed in. This means the user's callback currently has to handle the case where they're undefined, even though this will never happen. This PR changes the type signature of `onVisibleRegionChanged` to be more accurate to declare that these parameters will always be passed in.

(This change does not impact any existing users who are passing `onVisibleRegionChanged` callbacks that only take one parameter. I think the code marked the extra parameters as optional either because it was assumed it needed to be consistent with the VisibleRegion type (not sure if they should ideally be non-optional there too), or because of a mistaken belief it was necessary to in order to not break existing consumers passing in one-parameter callbacks, or because in the past these parameters were not always passed in.)


jassmith:
Can you please retarget this at 6.0.0

jassmith (approved):

jassmith:
For future reference this was not marked as such because I wasn't sure I wanted to continue to expose these props in this manner and well, Glide was the only user.

jassmith:
Will fix test failures in branch.

PR #828, opened 2023-12-21
Adjust selection to scroll dir

### LINKED ISSUE #819: Scroll Viewport when Active Selection Grows Out of View
This is a feature request.

When a user is modifying the active selection with their cursor, and the area of the selection grows off screen, the viewport should shift to keep it in view. This already exists when you grow a selection using arrow keys, and is the behavior in google sheets / excel.

We can maybe re-use `useAutoscroll` for this?

https://github.com/glideapps/glide-data-grid/blob/2ad08c86f43e78de6512ed4e1378494a2c5b97d3/packages/core/src/data-editor/data-editor.tsx#L2196

BrianHung:
Fixes #819 

Not 100% sure useEffect is the right way to adjust selection when scroll edge but seems to work.


packages/core/src/internal/data-grid/data-grid.tsx
```diff
@@ -510,10 +510,9 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
             const metaKey = ev?.metaKey === true;
             const isTouch = (ev !== undefined && !(ev instanceof MouseEvent)) || (ev as any)?.pointerType === "touch";
 
-            const edgeSize = 20;
             const scrollEdge: GridMouseEventArgs["scrollEdge"] = [
-                Math.abs(x) < edgeSize ? -1 : Math.abs(rect.width - x) < edgeSize ? 1 : 0,
-                Math.abs(y) < edgeSize ? -1 : Math.abs(rect.height - y) < edgeSize ? 1 : 0,
+                x < stickyX ? -1 : rect.width < x ? 1 : 0,
+                y < totalHeaderHeight ? -1 : rect.height < y ? 1 : 0,
```
jassmith, line 515: You're missing deps for this callback.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2091,6 +2091,59 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
 
     const [scrollDir, setScrollDir] = React.useState<GridMouseEventArgs["scrollEdge"]>();
 
+    // Adjust selection to include latest cell in visible region.
+    React.useEffect(() => {
```
jassmith, line 2095: I'm not sure this is right mechanism to do this. The easiest thing to do would be to find a way to reasonably re-emit the mouse move, triggering a new onItemHovered event as scroll is going on. The state update then would become a natural consequene of the artificial mousemove events.

jassmith (changes requested):

packages/core/src/data-editor/data-editor.tsx
BrianHung: Makes sense. Most of the logic for adjusting the selection in this useEffect I just took from the `adjustSelection` function, but emitting artificial mousemove events is a better way of code re-usability and correctness.

jassmith:
No pressure but I think this needs to change strategy if it can be merged. Let me know if you want me to keep this open or close it.

BrianHung:
Keep it open; looking into onItemHovered right now. Should it still be triggered using useEffect on scrollDir?

BrianHung:
Updated PR to use `onItemHoveredImpl` with `useEffect`. I wonder if it would even be easier if we modified `use-autoscroll` to accept a callback that's called every time the `scrollFn` is called. That would de-dupe the `useEffect` and `requestAnimationFrame`.

jassmith:
I mean I like less code, less to go wrong. Sounds like a good idea.

packages/core/src/data-editor/data-editor.tsx
BrianHung: PR now triggers onItemHovered by listening to the `useAutoscroll` callback.

jassmith:
Can you rebase on 6.0.0. I had some bugs when you branched.

jassmith:
nvm turns out I could

jassmith:
I tested this, works great. Just need to grok how you made this work now.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2447,8 +2446,11 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
         isActivelyDragging.current = false;
     }, []);
 
+    const hoveredRef = React.useRef<GridMouseEventArgs>();
     const onItemHoveredImpl = React.useCallback(
         (args: GridMouseEventArgs) => {
+            if (mouseEventArgsAreEqual(args, hoveredRef.current)) return;
```
jassmith, line 2452: This is here to prevent spurious events being sent out to the user. Confirm?

jassmith (approved):
This looks amazing. Great work. We may need to do some consolidation work at some point with computing the out of bounds cell location but great work.

packages/core/src/data-editor/data-editor.tsx
BrianHung: Yep; I put a console log statement there to check. It’s caused by the onItemHoveredImpl being called by onScroll.

BrianHung: I basically copied the same logic from data-grid since it also has hoveredRef and onItemHovered.

PR #829, opened 2023-12-23
Allow control of cell activation behavior

### LINKED ISSUE #687: onCellActivated event fires for single clicks even if cell is the same as previously clicked
**Issue:**

The onCellActivated event is triggered when a cell is clicked, even if it's the same cell as previously clicked, without considering any delay between clicks. This leads to confusion, as users may expect a double-click event instead of single-click behavior.

**Possible Cause:** 

`data-editor.tsx` contains the following code in the `onMouseUp` callback: 

```tsx
if (
  !isPrevented.current &&
  mouse?.previousSelection?.current?.cell !== undefined &&
  gridSelection.current !== undefined
) {
  const [selectedCol, selectedRow] = gridSelection.current.cell;
  const [prevCol, prevRow] = mouse.previousSelection.current.cell;
  if (col === selectedCol && col === prevCol && row === selectedRow && row === prevRow) {
      onCellActivated?.([col - rowMarkerOffset, row]);
      reselect(a.bounds, false);
      return true;
  }
}
```

**Suggested Solution:**

It would be helpful to either:

1. Pass additional arguments to the onCellActivated event as a second parameter, indicating how the event was triggered (e.g., with mouse event information).
2. Implement an extra check in the condition for handling double-click events specifically.

jassmith:
This PR enables control of the cell activation event and the opening of the overlay editor. Fixes #687

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -1922,7 +1922,6 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
     const lastMouseSelectLocation = React.useRef<readonly [number, number]>();
     const touchDownArgs = React.useRef(visibleRegion);
     const mouseDownData = React.useRef<{
-        wasDoubleClick: boolean;
         time: number;
         button: number;
         location: Item;
@@ -1937,9 +1936,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
             }
 
             const time = performance.now();
-            const wasDoubleClick = time - (mouseDownData.current?.time ?? -1000) < 250;
             mouseDownData.current = {
-                wasDoubleClick,
                 button: args.button,
                 time,
                 location: args.location,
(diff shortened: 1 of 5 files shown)


packages/core/src/internal/data-grid/data-grid.tsx
```diff
@@ -1034,6 +1038,11 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
                     ...args,
                     isLongTouch: true,
                 };
+            } else if (!args.isTouch && lastUpTimeValue !== 0 && Date.now() - lastUpTimeValue < 500) {
```
LukasMasuch, line 1041: If `double-click` is configured, cells cannot be activated on a touch device or?

LukasMasuch (approved):
LGTM 👍

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -651,6 +651,11 @@ export interface DataEditorProps extends Props, Pick<DataGridSearchProps, "image
      * Controls which directions fill is allowed in.
      */
     readonly allowedFillDirections?: FillHandleDirection;
+
+    /**
+     * Determins when a cell is considered activated
```
LukasMasuch, line 656: nit: Maybe add a bit more info to this comment about what cell activation actually means. Is it only about the cell overlay being opened?

packages/core/src/internal/data-grid/data-grid.tsx
jassmith: Good catch that is screwed up.

jassmith: Fixed

packages/core/src/data-editor/data-editor.tsx
jassmith: Resolved

PR #830, opened 2023-12-24
Allow specifying label and value independently in dropdown cell

### LINKED ISSUE #704: Allow to specify label and value in items for custom dropdown-cell
Looking at the code it has the facility but glide data grid only accepts a single string.
https://github.com/glideapps/glide-data-grid/blob/main/packages/cells/src/cells/dropdown-cell.tsx#L63

jassmith:
Fixes #704


LukasMasuch (approved):
LGTM 👍

PR #831, opened 2023-12-25
Make all keybindings rebindable

jassmith:


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -180,40 +180,170 @@ function shiftSelection(input: GridSelection, offset: number): GridSelection {
...
+    readonly selectRow: Keybind;
+    readonly selectColumn: Keybind;
 }
 
+type Keybinds = ConfigurableKeybinds & ForcedKeybinds;
+
+export type RealizedKeybinds = Readonly<Record<keyof ConfigurableKeybinds, string>> & ForcedKeybinds;
+
 const keybindingDefaults: Keybinds = {
-    selectAll: true,
```
jassmith, line 201: We should work on supporting these legacy binding setters as much as possible as well

BrianHung: https://github.com/ProseMirror/prosemirror-commands/blob/master/src/commands.ts#L744
https://github.com/ProseMirror/prosemirror-keymap/blob/master/src/keymap.ts

Might be helpful: ProseMirror splits their keyboard shortcuts into two concepts, keymaps and commands.

jassmith: Yeah it's worth considering but requires a more complex setup. I'd be curious what the use case would be.

PR #832, opened 2023-12-27
Make loading cell able to render a loading skeleton

jassmith:


packages/core/src/internal/data-grid/data-grid-types.ts
```diff
@@ -313,6 +313,8 @@ export interface BaseGridCell {
 /** @category Cells */
 export interface LoadingCell extends BaseGridCell {
     readonly kind: GridCellKind.Loading;
+    readonly skeletonWidth?: number;
+    readonly skeletonVariability?: number;
```
BrianHung, line 317: Think instead of `skeletonVariability`, `skeletonWidth` should either be a static `number` or a dynamic `(col: number, row: number) => number`, so it can left up to developer to be pseudo-random or deterministic.

BrianHung (approved):

packages/core/src/cells/loading-cell.tsx
```diff
@@ -1,13 +1,44 @@
...
+
+        let width = cell.skeletonWidth;
+        if (cell.skeletonVariability !== undefined && cell.skeletonVariability > 0) {
+            width += Math.round(getRandomNumber(col, row) * cell.skeletonVariability);
+        }
+
+        const hpad = theme.cellHorizontalPadding;
+        const rectHeight = 18;
+
+        roundedRect(ctx, rect.x + hpad, rect.y + (rect.height - rectHeight) / 2, width, rectHeight, 3);
```
LukasMasuch, line 39: Not really something to change in this PR, but I think the "roundness" could also be considered a theming aspect. Maybe it makes sense to introduce a single rounding radius theming option which is used as a default/fallback for rounded aspects like checkbox, image, drilldown, button, and tags cell.

packages/core/src/internal/data-grid/data-grid-types.ts
jassmith: The callback seems not needed as you get that information inside `getCellContent`. The variability is just there to make it easier to achieve semi-organic skeletons without writing your own randomizer. I just didn't want it to be the obvious fallback that people use Math.random() which is rather slow.

packages/core/src/cells/loading-cell.tsx
```diff
@@ -1,13 +1,44 @@
...
+            return;
+        }
+
+        let width = cell.skeletonWidth;
+        if (cell.skeletonVariability !== undefined && cell.skeletonVariability > 0) {
+            width += Math.round(getRandomNumber(col, row) * cell.skeletonVariability);
+        }
+
+        const hpad = theme.cellHorizontalPadding;
+        const rectHeight = 18;
```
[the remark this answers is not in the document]
jassmith: There wasn't a good option I could think of really. Using the cell height - 2*cellVerticalPadding is probably the most consistent method, but it is likely to look bad for some people. The only other things I could think to do were to make it configurable as a theme param, make it configurable as a cell param, or to make it parse the font param (which is a really bad idea).

packages/core/src/cells/loading-cell.tsx
jassmith: This is a good idea in general.

packages/core/src/cells/loading-cell.tsx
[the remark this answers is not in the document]
jassmith: I've attempting to strike a better balance in the recent commit

LukasMasuch (approved):
LGTM 👍

PR #835, opened 2023-12-28
Fix autoscroll bugs that cause wonkly behaviors

jassmith:


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2518,14 +2518,14 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
         let [col, row] = args.location;
         const visible = visibleRegionRef.current;
         if (xDir === -1) {
-            col = visible.x;
+            col = visible.extras?.freezeRegion?.x ?? visible.x;
```
jassmith, line 2521: This lets us account for forzen columns correctly.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2518,14 +2518,14 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
         } else if (xDir === 1) {
             col = visible.x + visible.width;
         }
         if (yDir === -1) {
             row = Math.max(0, visible.y);
         } else if (yDir === 1) {
-            row = Math.min(rows, visible.y + visible.height);
+            row = Math.min(rows - 1, visible.y + visible.height);
```
jassmith, line 2528: The last row is rows - 1

packages/core/src/internal/data-grid/data-grid.tsx
```diff
@@ -515,7 +515,7 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
             const isTouch = (ev !== undefined && !(ev instanceof MouseEvent)) || (ev as any)?.pointerType === "touch";
 
             const scrollEdge: GridMouseEventArgs["scrollEdge"] = [
-                x < stickyX ? -1 : rect.width < x ? 1 : 0,
+                x < 0 ? -1 : rect.width < x ? 1 : 0,
```
jassmith, line 518: Ideally this would know if the current selection is inside the freeze region or not, but since this is not currently known in this region of the code I didn't want to plumb it through. We could then conditionally scroll. For now however this will have to do.

packages/core/src/data-editor/data-editor.tsx
BrianHung: Makes sense; if you hover over frozen columns, you'd expect selection to also include it too.

BrianHung (approved):

packages/core/src/internal/data-grid/data-grid.tsx
jassmith: The problematic scenario here is that with `x < 0` you don't get autoscroll until you rmouse leaves the grid entirely. Ideally auto-scroll would start if you go over a sticky column and the current selected cell is not in the sticky region.

Issue #823, opened 2023-12-13
Group rows API


aleemstreak:
Hi there! 

We're from Streak (yc s11) and we're looking to move to Glide Data Grid. One missing feature from our current in-house implementation is the ability to have group rows.

Here's roughly what we mean by a group row (notice all the features are listed in the screenshot): 
<img width="1291" alt="Screenshot 2023-12-12 at 10 09 01 PM" src="https://github.com/glideapps/glide-data-grid/assets/1316303/3ea3a8a7-4633-4767-8d72-fcdd28c56cb9">

Here's a video of how the group row interactions feel wrt to scrolling and selection: https://www.loom.com/share/6cdaabf0bc3f402c90099749f2d65ada

Would love to start the discussion on how this might be implemented in GDG, we'd be happy to implement.

jassmith:
So the implementation of this is fairly straightforward:

- Make it possible to efficiently not draw vertical borders for certain rows without spanning the columns. This is going to make making your header easier.
- Add an API for marking a row as sticky
- When drawing the header texture, it also needs to draw the overflow region
- Update the input handler to understand these sticky rows when computing the click item.

I think we probably want to make this more of a first class feature and a feature you can build from a collection of features. The largest reason for this is the weird impact it has on input handling and sizing the header texture.

aleemstreak:
I'm not sure making it a regular row that hides the vertical seperators is going to work because we wouldn't want clipping in the row. For example, imagine the title of the group was really long:

<img width="1229" alt="Screenshot 2023-12-12 at 9 58 24 PM" src="https://github.com/glideapps/glide-data-grid/assets/1316303/aa849832-ba24-43f5-aced-7bc83b3116cc">

On the other hand, we'll prob need a way to figure out where the columns start and end in row because we might want to draw things in the group row that align with the headers like:

![summary](https://github.com/glideapps/glide-data-grid/assets/1316303/7527f7b3-4956-4ff0-9063-1615eeda1976)

aleemstreak:
On the other hand, we'll prob need a way to figure out where the columns start and end in row because we might want to draw things in the group row that align with the headers like:

![summary](https://github.com/glideapps/glide-data-grid/assets/1316303/7527f7b3-4956-4ff0-9063-1615eeda1976)

aleemstreak:
(and def agree on making this a first class feature because of the input handling)

jassmith:
API sketching:

```
interface StickyRowData {
    /**
[...]

PR #848, opened 2023-12-30
Freeze trailing row support

jassmith:
```[tasklist]
### Needed fixes
- [x] Fix rendering of highlight rings that span the freeze rows
- [x] The visibile region call needs to properly expose the frozen rows
- [x] Strict mode needs to be updated to allow `getCellContent` of these rows
- [x] Check ButtonCell animation works
- [x] Write a proper story
- [x] Fix renderStateProvider
- [x] When scrolled to bottom the grid line is double heavy (lo-dpi only) !needs confirmation of fix!
- [ ] Accessibility might need update
- [x] Scroll to does not properly account for frozen rows.
- [x] Issue with drawing of borders (see comment below)
- [x] Column selection doesn't include frozen rows
```


packages/core/src/internal/data-grid/data-grid-lib.ts
```diff
@@ -155,6 +155,18 @@ export function getStickyWidth(
...
 }
 
+export function getFreezeTrailingHeight(
+    rows: number,
+    freezeTrailingRows: number,
+    getRowHeight: number | ((row: number) => number)
+): number {
+    let result = 0;
+    for (let i = rows - freezeTrailingRows; i < rows; i++) {
+        result += typeof getRowHeight === "number" ? getRowHeight : getRowHeight(i);
```
jassmith, line 165: This is inefficient for large numbers of freeze regions, simple multiplication can work anytime the rowHeight is fixed which is the vast majority of cases.

packages/core/src/internal/data-grid/data-grid-lib.ts
```diff
@@ -224,18 +236,23 @@ export function getRowIndexForY(
...
 ): number | undefined {
     const totalHeaderHeight = headerHeight + groupHeaderHeight;
     if (hasGroups && targetY <= groupHeaderHeight) return -2;
     if (targetY <= totalHeaderHeight) return -1;
 
-    const lastRowHeight = typeof rowHeight === "number" ? rowHeight : rowHeight(rows - 1);
-    if (lastRowSticky && targetY > height - lastRowHeight) {
-        return rows - 1;
+    let y = height;
+    for (let fr = 0; fr < freezeTrailingRows; fr++) {
```
jassmith, line 246: Exactly how this loop works might need some comments

packages/core/src/internal/data-grid/data-grid-render.tsx
```diff
@@ -470,22 +468,33 @@ function drawGridLines(
         }
     }
 
-    const stickyHeight = getRowHeight(rows - 1);
-    const stickyRowY = height - stickyHeight + 0.5;
-    const lastRowSticky = trailingRowType === "sticky";
-    if (lastRowSticky) {
-        toDraw.push({ x1: minX, y1: stickyRowY, x2: maxX, y2: stickyRowY, color: hColor });
+    // here we need to push all the relevant lines for the sticky row
+    // const stickyHeight = getRowHeight(rows - 1);
```
jassmith, line 472: delete me

packages/core/src/internal/data-grid/data-grid-render.tsx
```diff
@@ -470,22 +468,33 @@ function drawGridLines(
...
-        const target = lastRowSticky ? height - stickyHeight : height;
+        const target = freezeY;
         while (y + translateY <= target) {
             const ty = y + translateY;
             // This shouldn't be needed it seems like... yet it is. We're not sure why.
-            if (ty >= minY && ty <= maxY - 1 && (!lastRowSticky || row !== rows - 1 || Math.abs(ty - stickyRowY) > 1)) {
+            if (
+                ty >= minY &&
+                ty <= maxY - 1 &&
+                (freezeTrailingRows === 0 || row !== rows - 1 || Math.abs(ty - freezeY) > 1)
```
jassmith, line 496: It is unclear at this time if this is working correctly.

packages/core/src/internal/data-grid/data-grid-render.tsx
```diff
@@ -1181,7 +1191,8 @@ function drawCells(
     ctx.font = font;
     const deprepArg = { ctx };
     const cellIndex: [number, number] = [0, 0];
-    const stickyRowHeight = trailingRowType === "sticky" ? getRowHeight(rows - 1) : 0;
+    const stickyRowHeight =
```
jassmith, line 1194: stickyRowHeight is a bad name now

packages/core/src/internal/data-grid/data-grid-render.tsx
```diff
@@ -2726,7 +2657,10 @@ export function drawGrid(arg: DrawGridArg, lastArg: DrawGridArg | undefined) {
             width: effectiveCols.length,
             height: lastRowDrawn - cellYOffset,
         },
-        freezeColumns
+        freezeColumns,
+        Array.from({ length: freezeTrailingRows })
```
jassmith, line 2661: This is probably inefficient

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2347,6 +2367,37 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                freezeColumns === 0
+                    ? undefined
+                    : {
+                          x: 0,
+                          y: region.y,
+                          width: freezeColumns,
+                          height: region.height,
+                      };
+
+            const freezeRegions: Rectangle[] = [];
```
LukasMasuch, line 2381: Does this theoretically support freezing any combination of cells?

jassmith: its setup so that we can make that happen

LukasMasuch:
From discord conversation: there seems to be a problem with rendering of borders, but not sure if this is specific to this branch:

https://github.com/glideapps/glide-data-grid/assets/2852129/e224a151-0415-4f4d-a514-dda4a2f87695

jassmith:
Fixed :)

LukasMasuch:
Another issue seems to be with column selection not including frozen rows:

https://github.com/glideapps/glide-data-grid/assets/2852129/a29cc690-8570-4c17-aa5a-5ca4cb02f9b9

jassmith:
Good catch!

LukasMasuch (approved):
Code looks good as far as I understood it 👍  

Two additional thoughts on this: 

1) maybe it would also benefit from a shadow like frozen columns and header row?
2) I think some of the uses for frozen rows will require to exclude frozen rows from some of the operations similar to the append row (cell highlighting, rowMarker, fill handle, ..). E.g. if it just shows some column statistics, it won't be really useful to select it together with the actual data above. This is also similar to [how aggrid handles pinned rows](https://www.ag-grid.com/javascript-data-grid/row-pinning/#non-supported-items). But that could be another config option to exclude those rows. I think this might be very similar to how grouped rows also need to be excluded from some operations as discussed in: https://github.com/glideapps/glide-data-grid/issues/823

Issue #840, opened 2023-12-29
Sparkline cell improvements


LukasMasuch:
Two small improvements for the sparkline cell:

1) The current version of `line` chart is usually called an `area` chart. It would be a nice addition if we had a proper line chart kind as well. And have the current version of `line` chart supported via the `area` graph kind. 
2) Add a configuration to hide the x-axis line.

Proposed API: 

```typescript
interface SparklineCellProps {
   // ... other props
    readonly graphKind?: "line" | "bar" | "area";
    readonly hideAxis?: boolean;
}
```

jassmith:
Lets ship these. Great improvements, simple to implement.

LukasMasuch:
Implemented by https://github.com/glideapps/glide-data-grid/pull/854

PR #854, opened 2024-01-06
Improve sparkline cell

LukasMasuch:
image

Closes https://github.com/glideapps/glide-data-grid/issues/840


jassmith (approved):
Nice nice. More charts more better.

Issue #839, opened 2023-12-29
Multi-select cell


LukasMasuch:
Similar to the dropdown cell, it would be a great addition to also have a multi-select cell as part of the `cells` package. E.g.:

<img width="385" alt="image" src="https://github.com/glideapps/glide-data-grid/assets/2852129/c1fa3d85-0f09-4a39-afa8-9a9d88b5e68e">

I think this should be pretty doable using `react-select` (also used for the `dropdown-cell`).

Proposed API:

```typescript
type SelectOption = string | { value: string; label: string };

interface MultiSelectCellProps {
    readonly kind: "multiselect-cell";
    readonly value: string[] | undefined | null;
    readonly allowedValues: readonly SelectOption[];
}
```

An additional feature could also be to allow adding values that are not part of the `allowedValues` ([react-select creatable docs](https://react-select.com/creatable)), e.g.:

```typescript
interface MultiSelectCellProps {
    // .. other props
    readonly creatable: boolean;
}
```

jassmith:
This is in effect what the TagsCell is. That said it could definitely use some cleanup.

LukasMasuch:
Oh indeed, tags cell is basically what is requested here. But I think it can really benefit from switching to the `react-select` multi-select input

jassmith:
Agreed

PR #859, opened 2024-01-08
Add multi select cell

LukasMasuch:
This PR adds a new multi-select cell type to the cell package:

image

It can be configured with the following properties:

```typescript
type SelectOption = { value: string; label?: string; color?: string };

interface MultiSelectCellProps {
    readonly kind: "multi-select-cell";
    /* The list of values of this cell. */
    readonly values: string[] | undefined | null;
    /* The list of possible options that can be selected.
    The options can be provided as a list of strings 
    or as a list of objects with the following properties:
    - value: The value of this option.
    - label: The label of this option. If not provided, the value will be used as the label.
    - color: The color of this option. If not provided, the default color will be used. */
    readonly options?: readonly (SelectOption | string)[];
    /* If true, users can create new values that are not part of the configured options. */
    readonly allowCreation?: boolean;
    /* If true, users can select the same value multiple times. */
    readonly allowDuplicates?: boolean;
}
```

The rounding of the tags can be configured via the `roundingRadius` theme option:

image

And the default color can be configured via the `bgBubble`, `bgBubbleSelected`, `textBubble` theme options.

Closes https://github.com/glideapps/glide-data-grid/issues/839

```[tasklist]
### Open Tasks
- [x] Use bubble text color if no other color was specified
```


LukasMasuch:
The feature will be finished in this branch: https://github.com/glideapps/glide-data-grid/pull/872

Issue #845, opened 2023-12-30
Unify rounding radii into theme property 


LukasMasuch:
Maybe it makes sense to introduce a single rounding radius theming option (e.g. `roundingRadius`) which is used as a default/fallback for rounded rects in GDG:

- Skeleton of loading cell (current value: 3)
- Checkbox (current value: 4)
- Image cell (current value: 4)
- Button cell (current value:  configurable, default is 0)
- Tags cell (current value: height / 2) 
- Bubble cell (current value: height / 2)
- Drilldown cell (current value: dynamic)

jassmith:
I like

jassmith:
This is now merged!

PR #861, opened 2024-01-09
Add rounding radius property to theme

LukasMasuch:
Introduces an optional `roundingRadius` theming option which is used as a default/fallback for rounded rects in GDG:

This is applied in: 

- Skeleton of loading cell
- Checkbox
- Image cell
- Button cell
- Tags cell
- Bubble cell
- Drilldown cell

Example with `roundingRadius==6`

image

Closes https://github.com/glideapps/glide-data-grid/issues/845

### DIFF
--- a/packages/core/src/cells/drilldown-cell.tsx
+++ b/packages/core/src/cells/drilldown-cell.tsx
@@ -41,7 +41,8 @@ const drilldownCache: {
 function getAndCacheDrilldownBorder(
     bgCell: string,
     border: string,
-    height: number
+    height: number,
+    rounding: number
 ): {
     el: HTMLCanvasElement;
     height: number;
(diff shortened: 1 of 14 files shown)


packages/core/src/internal/data-grid/data-grid-types.ts
```diff
@@ -372,6 +372,7 @@ export type ProvideEditorComponent<T extends InnerGridCell> = React.FunctionComp
     readonly target: Rectangle;
     readonly forceEditMode: boolean;
     readonly isValid?: boolean;
+    readonly theme: Theme;
```
LukasMasuch, line 375: I added this to make the theme available to custom editors, but eventually ended up not using this. But I think it could be useful to have access to this when implementing editors. Wdyt?

packages/core/src/cells/drilldown-cell.tsx
```diff
@@ -87,9 +87,8 @@ function getAndCacheDrilldownBorder(
 
     drilldownCache[key] = canvas;
 
-    const trueRounding = Math.min(rounding, targetWidth / 2, targetHeight / 2);
```
LukasMasuch, line 90: I removed this because the overlay editor also just used a static value (6px), but I can also try to figure out how to keep this logic in.

jassmith (approved):
LGTM

PR #862, opened 2024-01-11
Use `accentLight` color for resize indicator

LukasMasuch:
Uses `accentLight` color for resize indicator instead of `accentColor`:

https://github.com/glideapps/glide-data-grid/assets/2852129/f9c759ec-7d7c-404c-8704-a0aaae0e3887

### DIFF
--- a/packages/core/src/internal/data-grid/data-grid-render.tsx
+++ b/packages/core/src/internal/data-grid/data-grid-render.tsx
@@ -1828,7 +1828,7 @@ function drawColumnResizeOutline(
     ctx.lineTo(yOffset, height);
 
     ctx.lineWidth = 2;
-    ctx.strokeStyle = theme.accentColor;
+    ctx.strokeStyle = theme.accentLight;
 
     ctx.stroke();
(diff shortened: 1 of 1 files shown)


packages/core/src/internal/data-grid/data-grid-render.tsx
```diff
@@ -1828,7 +1828,7 @@ function drawColumnResizeOutline(
     ctx.lineTo(yOffset, height);
 
     ctx.lineWidth = 2;
-    ctx.strokeStyle = theme.accentColor;
+    ctx.strokeStyle = theme.accentLight;
```
jassmith, line 1831: we should probably blend this onto `bgCell` so that it is a solid color

jassmith (changes requested):

packages/core/src/internal/data-grid/data-grid-render.tsx
LukasMasuch: Added a blend 👍

PR #865, opened 2024-01-12
Apply visual fixes to cells

LukasMasuch:
1. Fix font color when clicking an item in the dropdown cell menu:

Before:

image

After:
image

2. Use pointer cursor for dropdown cell menu
3. Use the defined cell padding for the dropdown cell menu
4. Fix the measurement of some cells that use static values instead of the defined cell padding.


jassmith (approved):
Good cleanups!

PR #872, opened 2024-01-15
Add multi-select cell

LukasMasuch:
This PR adds a new multi-select cell type to the cell package:

image

It can be configured with the following properties:

```typescript
type SelectOption = { value: string; label?: string; color?: string };

interface MultiSelectCellProps {
    readonly kind: "multi-select-cell";
    /* The list of values of this cell. */
    readonly values: string[] | undefined | null;
    /* The list of possible options that can be selected.
    The options can be provided as a list of strings 
    or as a list of objects with the following properties:
    - value: The value of this option.
    - label: The label of this option. If not provided, the value will be used as the label.
    - color: The color of this option. If not provided, the default color will be used. */
    readonly options?: readonly (SelectOption | string)[];
    /* If true, users can create new values that are not part of the configured options. */
    readonly allowCreation?: boolean;
    /* If true, users can select the same value multiple times. */
    readonly allowDuplicates?: boolean;
}
```

The rounding of the tags can be configured via the `roundingRadius` theme option:

image

And the default colors can be configured via the `bgBubble`, `bgBubbleSelected`, and `textBubble` theme options.

Closes https://github.com/glideapps/glide-data-grid/issues/839

### DIFF
--- a/packages/cells/package.json
+++ b/packages/cells/package.json
@@ -54,10 +54,12 @@
         "@linaria/react": "^4.5.3",
         "@toast-ui/editor": "3.1.10",
         "@toast-ui/react-editor": "3.1.10",
+        "chroma-js": "^2.4.2",
         "react-select": "^5.2.2"
     },
     "devDependencies": {
         "@babel/cli": "^7.16.0",
+        "@types/chroma-js": "^2.4.3",
         "@types/prosemirror-commands": "^1.0.4",
         "@types/react": "16.14.21",
         "eslint": "^8.19.0",
--- a/packages/cells/package.json
+++ b/packages/cells/package.json
@@ -54,12 +54,10 @@
         "@linaria/react": "^4.5.3",
         "@toast-ui/editor": "3.1.10",
         "@toast-ui/react-editor": "3.1.10",
-        "chroma-js": "^2.4.2",
         "react-select": "^5.2.2"
     },
     "devDependencies": {
         "@babel/cli": "^7.16.0",
-        "@types/chroma-js": "^2.4.3",
         "@types/prosemirror-commands": "^1.0.4",
         "@types/react": "16.14.21",
         "eslint": "^8.19.0",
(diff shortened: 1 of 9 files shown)


packages/cells/package.json
```diff
@@ -54,10 +54,12 @@
         "@linaria/react": "^4.5.3",
         "@toast-ui/editor": "3.1.10",
         "@toast-ui/react-editor": "3.1.10",
+        "chroma-js": "^2.4.2",
```
jassmith, line 57: This dep is almost 1/3rd the total weight of GDG on its own. It looks like we want to use it for parsing, `brighten`, `darken`, `luminance`, and formatting.

`luminance` is an extremely simple calculation: https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef

`brighten` is just darken with an inverted sign

`darken` is in fact the only complicated operator here because chroma-js will use the CIELAB colorspace to perform this computation. Conversion to/from lab is pretty expensive, but more importantly just takes a fair bit of code. Implementing an sRGB version of this would be the only tradeoff required and may produce results with sufficient fidelity to serve the purpose.

Converting to hex/css is a trivial operator.

LukasMasuch: Sounds good. `darken` and `brighten` are not even really needed. I removed the usage of chroma-js and added the `getLuminance` and `toHex` from color2k to the `color-parser`

packages/cells/src/cells/multi-select-cell.tsx
```diff
@@ -0,0 +1,514 @@
...
+        let { x } = drawArea;
+        let row = 1;
+
+        let y =
+            rows === 1
+                ? drawArea.y + (drawArea.height - BUBBLE_HEIGHT) / 2
+                : drawArea.y + (drawArea.height - rows * BUBBLE_HEIGHT - (rows - 1) * BUBBLE_PADDING) / 2;
+        for (const value of values) {
+            const matchedOption = options.find(t => t.value === value);
+            const colorHex = matchedOption?.color
```
jassmith, line 413: Why does this need to be hex? It seems like it could be any css color string?

LukasMasuch: No need for this, removed it 👍

PR #880, opened 2024-01-24
Fix row theme and column theme cutting off when entering into blank region of Grid

jassmith:
CleanShot 2024-01-23 at 22 33 13@2x

Previously everything below the last data row would just be white

### DIFF
--- a/packages/core/src/common/styles.ts
+++ b/packages/core/src/common/styles.ts
@@ -1,4 +1,5 @@
 import React from "react";
+import { blend } from "../internal/data-grid/color-parser.js";
 
 // theme variable precidence
 
@@ -159,7 +160,11 @@ export function mergeAndRealizeTheme(theme: Theme, ...overlays: Partial<Theme | 
             for (const key in overlay) {
                 // eslint-disable-next-line no-prototype-builtins
                 if (overlay.hasOwnProperty(key)) {
-                    merged[key] = (overlay as any)[key];
+                    if (key === "bgCell") {
+                        merged[key] = blend(overlay[key] as string, merged[key]);
+                    } else {
+                        merged[key] = (overlay as any)[key];
+                    }
                 }
             }
         }
--- a/packages/core/src/docs/examples/theme-per-row.stories.tsx
+++ b/packages/core/src/docs/examples/theme-per-row.stories.tsx
@@ -35,27 +35,39 @@ export default {
 export const ThemePerRow: React.VFC = () => {
     const { cols, getCellContent, onColumnResize, setCellValue } = useMockDataGenerator(5);
 
+    const realCols = React.useMemo(() => {
+        const c = [...cols];
+        c[3] = {
+            ...c[3],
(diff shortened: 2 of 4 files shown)


packages/core/src/common/styles.ts
```diff
@@ -159,7 +160,11 @@ export function mergeAndRealizeTheme(theme: Theme, ...overlays: Partial<Theme |
             for (const key in overlay) {
                 // eslint-disable-next-line no-prototype-builtins
                 if (overlay.hasOwnProperty(key)) {
-                    merged[key] = (overlay as any)[key];
+                    if (key === "bgCell") {
+                        merged[key] = blend(overlay[key] as string, merged[key]);
```
jassmith, line 164: Without doing this we retain the previous requirement that bgCell always be a solid color. Now only the main theme need be a solid color.

packages/core/src/docs/examples/theme-per-row.stories.tsx
```diff
@@ -35,27 +35,39 @@ export default {
...
-                          bgCell: "#f0f8ff",
-                          borderColor: "#3f90e0",
+                          bgCell: "#e0f0ff88",
+                          //   borderColor: "#3f90e0",
                       }
             }
             onCellEdited={setCellValue}
             onColumnResize={onColumnResize}
-            rows={1_000_000}
+            rows={10}
```
jassmith, line 70: This makes it easier to see the change

packages/core/src/internal/data-grid/render/data-grid-render.lines.ts
```diff
@@ -146,6 +146,134 @@ export function overdrawStickyBoundaries(
...
+        }
+    }
+
+    if (toDraw.length === 0) return;
+
+    let color: string | undefined;
+    ctx.beginPath();
+    // render in reverse order because we computed and added the columns last, but they should actually be lower
+    // priority than the rows.
+    for (let i = toDraw.length - 1; i >= 0; i--) {
```
jassmith, line 258: The complexity here is just to reduce the number of draw calls. Fewer draw calls === faster

BrianHung:
image

Seems to be a missing row when trailingRowOptions and onRowAppended are defined.

BrianHung (changes requested):

packages/core/src/internal/data-grid/render/data-grid-render.lines.ts
```diff
@@ -146,6 +146,134 @@ export function overdrawStickyBoundaries(
...
+                    y: ty,
+                    w: maxX - minX,
+                    h: rh,
+                    color: rowThemeBgCell,
+                });
+            }
+        }
+
+        y += rh;
+        if (row < rows) y2 = y;
```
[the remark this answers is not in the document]
jassmith: Thank you! Great catch.

BrianHung:
LGTM after that fix :)

PR #882, opened 2024-01-24
dont emit onItemHovered if resizing or reordering

### LINKED ISSUE #868: Don't trigger `onItemHovered` if column is being resized
The `onItemHovered` callback gets triggered during column resizing. I think this is a bit unexpected since a user probably doesn't intend to hover a cell/column header during the column resizing.

BrianHung:
Fixes #868


jassmith:
Great fix

PR #892, opened 2024-02-01
Enforce readonly mode for `uri-cell`

LukasMasuch:
If the `uri-cell` is configured as `readonly`, there shouldn't be any way that it still allows editing (this is currently the case).

### DIFF
--- a/packages/core/src/cells/uri-cell.tsx
+++ b/packages/core/src/cells/uri-cell.tsx
@@ -115,7 +115,10 @@ export const uriCellRenderer: InternalCellRenderer<UriCell> = {
         const { onChange, value, forceEditMode, validatedSelection } = p;
         return (
             <UriOverlayEditor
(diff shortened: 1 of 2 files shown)


packages/core/src/cells/uri-cell.tsx
```diff
@@ -115,7 +115,10 @@ export const uriCellRenderer: InternalCellRenderer<UriCell> = {
         const { onChange, value, forceEditMode, validatedSelection } = p;
         return (
             <UriOverlayEditor
-                forceEditMode={forceEditMode || (cell.hoverEffect === true && cell.onClickUri !== undefined)}
+                forceEditMode={
+                    value.readonly !== true &&
```
LukasMasuch, line 119: Not sure about this check. It is not really necessary since `readonly` will be anyways enforced within `UriOverlayEditor``

PR #896, opened 2024-02-03
Add support for indicator icons (normally used for showing sort)

jassmith:
- Add support for header indicator icons
- Fix test


fabio-carvalho88:
hi @jassmith is this what i need to achieve someting like this? At the moment I can't use it

version:     `@glideapps/glide-data-grid": "^6.0.3`


image

PR #898, opened 2024-02-07
Row Grouping

jassmith:
CleanShot 2024-02-07 at 21 10 46@2x

Adds row grouping support. This is not a complete implementation yet as it does not yet support overriding vertical borders or implement sticky headers.


aleemstreak:
Wow thanks. Sorry we haven't gotten to it yet (we were planning for it in the next month or so)

jassmith:
awesome! Please let me know if you give it a run!

PR #908, opened 2024-02-26
Take cell padding into account when measuring column title width

sfc-gh-jipark:
Small fix for taking `theme.cellHorizontalPadding` into account (like how most [cells](https://github.com/sfc-gh-jipark/glide-data-grid/blob/cf6c69a7f79427adb69e88bb099ae1e43b87e706/packages/core/src/cells/text-cell.tsx#L59) are being measured) when measuring column width based on column title in `use-column-sizer`. Currently, it's based on hardcoded 16px (twice the default 8px cellHorizontalPadding) so measurement is inaccurate when custom theme is configured

### DIFF
--- a/packages/core/src/data-editor/use-column-sizer.ts
+++ b/packages/core/src/data-editor/use-column-sizer.ts
@@ -62,7 +62,7 @@ export function measureColumn(
             }
         }
     }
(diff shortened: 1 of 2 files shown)


LukasMasuch (approved):
LGTM 👍

jassmith (approved):

PR #915, opened 2024-03-09
Keep focus in overlay editor

### LINKED ISSUE #910: Escape doesn't close cell overlay consistently
If you open a cell and press Escape, it closes the cell overlay. But if you click inside the cell, escape will do nothing:

https://github.com/glideapps/glide-data-grid/assets/2852129/23898488-2505-4f7f-a1c4-5da6c1b145fd

Is there a reason for this? It would probably be more consistent if `esc` always closes the cell overlay.

BrianHung:
Tries to fix #910 

We should add some tests to check that click within cell and then pressing escape works for each internal cell editor.


BrianHung:
Not a blocker for me; it is important that we find a way to always maintain focus if user is interacting with the grid, and not have it be lost unexpectedly to body. There can be other alternatives to this PR as well.

PR #924, opened 2024-03-15
Improve Performance in Row Grouping

bkorobeinikov:
There is a performance bottleneck when using the row grouping feature. It is noticeable with a large number of rows and about >~20 groups or so.

Profiler showed that the call to `mapRowIndexToPath` during row height resolution takes most of the time, and it gets slower scrolling towards the end of the list. 

`rowIndex` property was added to identify the exact position of the group in the list. This allowed to create a map that then could be looked up to identify whether the row is a group row or not and resolve the height accordingly, avoiding the expensive call to `mapRowIndexToPath`.

For further performance improvements, with `rowIndex` property functions like `rowNumberMapperOut` and `mapRowIndexToPath` could be changed to do a binary search, because `mapRowIndexToPath` can be used as `mapper` function in the `getCellContent` consumer code.

### scrolling profiler recordings:

before change
image

after
image

related to https://github.com/glideapps/glide-data-grid/issues/823

### DIFF
--- a/packages/core/src/data-editor/row-grouping.ts
+++ b/packages/core/src/data-editor/row-grouping.ts
@@ -4,6 +4,10 @@ import type { DataEditorProps } from "./data-editor.js";
 import type { DataGridProps } from "../internal/data-grid/data-grid.js";
 import { whenDefined } from "../common/utils.js";
 
+type Mutable<T> = {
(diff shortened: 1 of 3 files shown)


theiliad:
@jassmith I'm curious if this has been reviewed? I'm also having issues with large resource clogs when using row grouping

PR #927, opened 2024-03-20
Support group header hover and regular bg

rferreira98:
With this PR we'll be able to customize background colors for group headers and also change the hover color as well without affecting regular headers.


hakobyanheghine:
Hey @jassmith ,

When can we get this merged? This would be very helpful for my case as well! 

Thanks!

sergeynersesyan:
This is very helpful feature. Waiting for it to be merged

Issue #934, opened 2024-04-02
Row marker header checkbox is incorrectly shown for single row selection


LukasMasuch:
Configuring `rowSelect` as `single` still shows a not-functional checkbox in the row marker header. I would expect that this isn't shown since it this feature doesn't really makes sense for `single` row selection. 

https://github.com/glideapps/glide-data-grid/assets/2852129/7f6d9a94-a0a4-4e34-863d-ed77dbdaa407

PR #935, opened 2024-04-03
Disable header row marker on single selection

LukasMasuch:
Configuring `rowSelect` as `single` still shows a not-functional checkbox in the row marker header. I would expect that this isn't shown since this feature doesn't really make sense for single-row selection. This PR disables the row marker header for `rowSelect` == `single`.

https://github.com/glideapps/glide-data-grid/assets/2852129/7f6d9a94-a0a4-4e34-863d-ed77dbdaa407

Closes https://github.com/glideapps/glide-data-grid/issues/934

In addition, in the current impl. it was also possible to have multi-selection even if `row select` is set to `single.` For example, touch devices would falsely allow multi-selection. This PR also enforces the `single` configuration.

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -877,6 +877,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
     const rowMarkerTheme = rowMarkersObj?.theme ?? p.rowMarkerTheme;
     const headerRowMarkerTheme = rowMarkersObj?.headerTheme;
     const headerRowMarkerAlwaysVisible = rowMarkersObj?.headerAlwaysVisible;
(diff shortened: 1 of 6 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -1795,7 +1798,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                         } else {
                             setSelectedRows(CompactSelection.fromSingleSelection(newSlice), undefined, isMultiRow);
                         }
-                    } else if (isMultiRow || args.isTouch || rowSelectionMode === "multi") {
+                    } else if (rowSelect === "multi" && (isMultiRow || args.isTouch || rowSelectionMode === "multi")) {
```
LukasMasuch, line 1801: An additional check for `rowSelect` == `multi` here to make sure that multi-selection is only allowed if it is actually configured.

PR #948, opened 2024-04-29
Use Range type instead of Item type

willium:
Item type is not semantically correct here. An Item is a [row, col] pair, but this callback argument should be a [startIndex, endIndex] pair.


lukasmasuch (approved):
LGTM 👍

PR #961, opened 2024-06-10
getMouseArgsForPosition ref API

BrianHung:


BrianHung:
@lukasmasuch I use it for spreadsheets! getMouseArgsForPosition lets use know what cell or row or column we're clicking or pointing over. We can then convert those positions into A1 notation or A:A.

Example of that below in Google Sheets.

https://github.com/user-attachments/assets/aa747903-7f0d-43a0-b2ef-4f356b8e7351

lukasmasuch (approved):
Ok, sounds good 👍 But maybe two more things to add:

Add a documentation section to `API.md` with a bit more info (same exists for the other ref methods, [example](https://github.com/glideapps/glide-data-grid/blob/478c7ec7007fe5c2a04a958640b9a64001e28d54/packages/core/API.md#updatecells)). And maybe a storybook case could be a nice to have in case you can think of a nice case to showcase this.

PR #963, opened 2024-06-11
onColumnProposeMove should prevent drag

### LINKED ISSUE #945: onColumnProposeMove doesn't prevent actual column move
When using onColumnProposeMove to return false for a given column move, graphically when dragging we see that the columns indeed do not automatically swap places when they otherwise would, however when the mouse up happens and the moved column is "dropped", the column switch still occurs anyway.

An example of this problem can be seen in the Rearrange Columns story, where a function onColProposeMove checks if the endIndex is not 3. If you go to the story and drag the other columns, they instantly swap places when dragged, but column 3 does not, however it still will swap places after you mouse up to release the dragged column.

BrianHung:
Fixes #945


lukasmasuch (approved):
LGTM 👍

PR #964, opened 2024-06-11
onKeyDown rowMarkerOffset

### LINKED ISSUE #938: `onKeyDown` gives colum index 1 instead of 0
### Issue: 
When clicking a row 1 column 1 cell (data cell, not header nor. rowmaker column), the index of column is not updating to 0. So when press a keydown, it gives back column index = 1.

```javascript
onKeyDown={e => {
  const [coldix, rowidx] = e.location
  // coldix == 1
}}
```

---

I suspect it is `handleMaybeClick` on `/node_modules/@glideapps/glide-data-grid/src/data-editor/data-editor.tsx`, in the `if (isValidClick)` block it should `setGridSelection` to `expectedExternalGridSelection`.

Can you confirm this is a bug, or it is expected?
Thank you for your great library.

BrianHung:
Fixes #938


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3321,7 +3321,10 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
             let cancelled = false;
             if (onKeyDownIn !== undefined) {
                 onKeyDownIn({
-                    ...event,
+                    ...event, // FIXME: Mutate
```
lukasmasuch, line 3324: Can the `FIXME: Mutate` comment be removed?

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3321,7 +3321,10 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                    ...event.location && {
+                        location: [event.location[0] - rowMarkerOffset, event.location[1]] as any,
```
lukasmasuch, line 3326: nit: maybe add a comment here in code why this needs to be adjusted

packages/core/src/data-editor/data-editor.tsx
BrianHung: done!

lukasmasuch (approved):
LGTM 👍

PR #969, opened 2024-06-23
expose renderers prop to allow custom internal renderers

BrianHung:


lukasmasuch (approved):
LGTM 👍 Might be good to also list this in the API.md and maybe add one example to the custom drawing story or a new story.

PR #981, opened 2024-07-12
add header disabled to row markers

### LINKED ISSUE #940: Add ability to remove checkbox in rowMarkers header
I like rowMarkers for row selection and the visual index, but I don't want users to be able to select all rows/columns. It'd be great to add a prop to rowMarkers config to disable this, if it's not too difficult!

BrianHung:
Closes #940


packages/core/src/internal/data-grid/data-grid.tsx
```diff
@@ -940,6 +940,7 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
         : headerHovered || clickableInnerCellHovered || editableBoolHovered || groupHeaderHovered
         ? "pointer"
         : "default";
+    console.log("CURSOR", cursor, cursorOverride, headerHovered);
```
lukasmasuch, line 943: needs to be removed

packages/core/src/internal/data-grid/render/data-grid-render.header.ts
```diff
@@ -77,7 +77,7 @@ export function drawGridHeaders(
             ctx.font = theme.baseFontFull;
         }
         const selected = selection.columns.hasIndex(c.sourceIndex);
-        const noHover = dragAndDropState !== undefined || isResizing;
+        const noHover = dragAndDropState !== undefined || isResizing || c.headerRowMarkerDisabled === true;
```
lukasmasuch, line 80: Same as the comment above. Not sure if this is event required since the exising logic of `headerRowMarkerDisabled` should already prevent the hover checkbox.

packages/core/src/internal/data-grid/data-grid.tsx
```diff
@@ -915,7 +915,7 @@ const DataGrid: React.ForwardRefRenderFunction<DataGridRef, DataGridProps> = (p,
     const [overFill, setOverFill] = React.useState(false);
 
     const [hCol, hRow] = hoveredItem ?? [];
-    const headerHovered = hCol !== undefined && hRow === -1;
+    const headerHovered = hCol !== undefined && hRow === -1 && columns[hCol].headerRowMarkerDisabled !== true;
```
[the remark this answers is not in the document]
BrianHung: headerRowMarkerDisabled hides the row marker checkbox, but the header cell still receives hover styles, shows a pointer cursor, and triggers select-all behavior on click.

packages/core/src/internal/data-grid/render/data-grid-render.header.ts
BrianHung: Only hides but doesn't prevent interaction; see comment above.

PR #1004, opened 2024-11-05
Add ability to set scroll behavior when calling ref.scrollTo

Tintef:
### What?

- Accept `behavior` as part of the `options` object when calling `scrollTo` method
- Accept `behavior` when calling `appendRow`

### Why?

So we can have a smooth transition when programmatically adding a row or scrolling to a given cell.

The current way to do this, as far as I know, is adding a class with `scroll-behavior: smooth` to the editor, which causes arrow navigation to also have the smooth behavior which feels awful :/


lukasmasuch (approved):
LGTM 👍

PR #1005, opened 2024-11-13
fix: highlight border to be drawn many times in different places in the frozen area

starlight-93:
image

There are multiple cropping regions, resulting in the border not being cropped


lukasmasuch (approved):
LGTM 👍

Issue #1022, opened 2025-02-14
Datagrid events are blocked when used inside a constrained layer like Modal


sfc-gh-yalu:
We have a use case where we show a datagrid within a modal dialog, where events outside of the dialog is disabled for trapping focus. For example, user won't (and shouldn't) be able to click on a button "beneath" (outside) the dialog backdrop when the dialog is open. However, we should still allow interactions on the glide datagrid (as part of dialog content) like resize, select and edit.

Due to the current glide [implementation](https://github.com/glideapps/glide-data-grid/blob/5983dcabd2fb55b675009813709752008da6d424/packages/core/src/internal/data-grid/data-grid.tsx#L1098) only attaching the event handlers on `window`, the event triggered from a datagrid within the above modal will get blocked and won't reach up there, causing these interactive features to stop working.

Can we have a way to customize the element where event handlers are attached (e.g. to the datagrid [container](https://github.com/glideapps/glide-data-grid/blob/5983dcabd2fb55b675009813709752008da6d424/packages/core/src/internal/data-editor-container/data-grid-container.tsx#L44) instead of `window`) to support this use case?

PR #1025, opened 2025-03-06
Add experimental support for custom event target

lukasmasuch:
Implements support for a custom event target in the data grid to allow event handling on a specified DOM element rather than on `window`. This can be set via `windowEventTarget` as part of the `experimental` properties.

Closes https://github.com/glideapps/glide-data-grid/issues/1022

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,33 @@
+const path = require("path");
+
+module.exports = {
+    stories: ["../**/src/**/*.stories.tsx"],
+    addons: [getAbsolutePath("@storybook/addon-storysource"), getAbsolutePath("@storybook/addon-controls")],
+
+    typescript: {
+        reactDocgen: false,
+    },
+
+    async viteFinal(config) {
+        // We need to dynamically import these since they use ESM
+        const { mergeConfig } = await import("vite");
+        const linaria = (await import("@linaria/vite")).default;
+
+        return mergeConfig(config, {
+            plugins: [linaria()],
+        });
+    },
+
+    framework: {
+        name: getAbsolutePath("@storybook/react-vite"),
+        options: {},
+    },
+
+    docs: {
+        autodocs: false,
+    },
+};
+
+function getAbsolutePath(value) {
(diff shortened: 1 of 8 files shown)


setup-react-18-test.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-npm i -D react@latest react-dom@latest @testing-library/react@latest @testing-library/react-hooks@latest @testing-library/user-event@14.5.1 react-test-renderer@latest
\ No newline at end of file
+npm i -D react@latest react-dom@latest @testing-library/react@latest @testing-library/react-hooks@latest @testing-library/user-event@14.5.1 react-test-renderer@latest @testing-library/dom
```
lukasmasuch, line 3: Needed to fix the tests.

.storybook/main.js
lukasmasuch, line 1: Added to have Storybook working again for local usage. There might be a different workaround, but seems like it doesn't hurt to have the `main.js` in addition to the `main.ts`.

BrianHung (approved):
Overall, LGTM!

One concern I have playing around with the storybook, is what should be the behavior of click outside when the cell editor is open?

https://github.com/glideapps/glide-data-grid/blob/main/packages/core/src/internal/click-outside-container/click-outside-container.tsx

End-users can also use the `isOutsideClick` prop, but I'm thinking we should be consistent with defaults and pass down `eventTarget` to the click-outside-container: we should close the editor on outside click when user tries to click outside the event target.

lukasmasuch:
> End-users can also use the isOutsideClick prop, but I'm thinking we should be consistent with defaults and pass down eventTarget to the click-outside-container: we should close the editor on outside click when user tries to click outside the event target.

Thanks, that's a good point! I played around a bit with it and decided to pass down the event target to the click-outside-container as well. However, in the scenario of the storybook where the clicks on window are fully consumed/blocked, there doesn't seem to be a good way to close the editor for clicks that are outside of the provided event target. I think the user has to define a good event target that makes sense for a given usage scenario.

PR #1032, opened 2025-05-10
Update exported drawTextCell to handle wrapping.

infalmo:


lukasmasuch:
LGTM 👍  @BrianHung do you see any reason why not to expose these options in the exported function?

BrianHung:
Nope, LGTM!

PR #1033, opened 2025-05-11
Fix incorrect calculation of DataEditor height.

infalmo:


BrianHung:
I believe sampling only 10 rows as opposed to _all rows_ was an intentional performance optimization, as summing heights for thousands or millions of rows is an issue.

lukasmasuch:
Yep, I think we can close this since its an intended optimisation. But maybe we could use a slightly higher value, e.g. 50 or 100 rows, to get a better approximation without adding to much computation overhead.

PR #1038, opened 2025-06-05
feat: provideEditorCallback access cell location

BrianHung:
Most of api is 

`(cell: Item, value: GridCell)` but that would be a breaking change here.


lukasmasuch (commented):
This looks reasonable, but its also a breaking change, right? All uses of `provideEditor` would need to be adapted to accept the new parameter?

BrianHung:
Yeah, it is a breaking change. Maybe we can pass in (cell: GridCell & Item), then it wouldn't be a breaking change with a required second arg. What do you think?

lukasmasuch:
LGTM 👍 I think the main downside with the updated impl. is that it would overwrite a `location` prop of a built-in cell. But none of the built-in cells use this property yet -> should be good to go 

One small related nice-to-have addition might be a storybook case for `provideEditor`. I don't think we have one at the moment. But we can also do that in another PR, whatever you prefer.

PR #1040, opened 2025-06-13
fix: use correct font to measure column title width

murrayju:
The `measureColumn` code is incorrectly using the `baseFontFull` when measuring the width of the title (header), when `headerFontFull` should be used instead.

When the header is wider than the data, and therefore determines the auto-sized width, the computed width can be too narrow, cutting off the right padding and some of the text. This is assuming the header font is larger/bolder than the base font. Could have the inverse problem (auto-size too wide), as well.

Before:
image

After:
image

### DIFF
--- a/packages/core/src/data-editor/use-column-sizer.ts
+++ b/packages/core/src/data-editor/use-column-sizer.ts
@@ -62,7 +62,9 @@ export function measureColumn(
             }
         }
     }
+    ctx.font = theme.headerFontFull;
     max = Math.max(max, ctx.measureText(c.title).width + theme.cellHorizontalPadding * 2 + (c.icon === undefined ? 0 : 28));
+    ctx.font = theme.baseFontFull;
     const final = Math.max(Math.ceil(minColumnWidth), Math.min(Math.floor(maxColumnWidth), Math.ceil(max)));
 
     return {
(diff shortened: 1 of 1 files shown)


packages/core/src/data-editor/use-column-sizer.ts
```diff
@@ -62,7 +62,9 @@ export function measureColumn(
             }
         }
     }
+    ctx.font = theme.headerFontFull;
     max = Math.max(max, ctx.measureText(c.title).width + theme.cellHorizontalPadding * 2 + (c.icon === undefined ? 0 : 28));
+    ctx.font = theme.baseFontFull;
```
lukasmasuch, line 67: Maybe it's slightly better to save the current `ctx.font` into a variable and reset it from that:

```typescript
const currentFont = ctx.font;
ctx.font = theme.headerFontFull;
max = Math.max(max, ctx.measureText(c.title).width + theme.cellHorizontalPadding * 2 + (c.icon === undefined ? 0 : 28));
ctx.font = currentFont;
```

murrayju: Yeah, I had considered this as well. I think either could make sense in the existing code. I find it a bit odd that the caller sets `ctx.font` (for the base font), vs this function controlling that itself, but was avoiding making broader changes.

Anyway, pushed your suggestion:
https://github.com/glideapps/glide-data-grid/pull/1040/commits/62df81dfc0429d56bf9df4ca2eaf31db6927bd67

PR #1041, opened 2025-06-17
Update minor dependency versions

lukasmasuch:
Updates all the minor dependency versions. I will attempt to do some major version updates in a follow-up PR.


BrianHung (approved):
LGTM; updates package.json and passes CI!

PR #1043, opened 2025-06-18
Adapt range cell rendering based on the configured font size

lukasmasuch:
Changes the range cell rendering to be adapted based on the configured font. Instead of a static `12px` font size and `6px` range height, we take 90% of the configure font size for the range label and use 50% of the font (em) height for the range visualization. This makes the range cell better adapted to the configured theme.


BrianHung (approved):
LGTM; refactors range cell to remove fixed constants.

PR #1045, opened 2025-06-18
Allow theming of bubble dimensions

lukasmasuch:
Adds three new theming parameters `bubbleHeight`, `bubblePadding`, and `bubbleMargin` that allow configuring the bubble dimensions via the theme. Bubbles are used within multiple cells (e.g. bubble cell, multiselect cell, tags cell).

### DIFF
--- a/packages/cells/src/cells/multi-select-cell.tsx
+++ b/packages/cells/src/cells/multi-select-cell.tsx
@@ -36,9 +36,6 @@ interface MultiSelectCellProps {
     readonly allowDuplicates?: boolean;
 }
 
-const BUBBLE_HEIGHT = 20;
-const BUBBLE_PADDING = 6;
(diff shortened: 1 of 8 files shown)


BrianHung (approved):
Extracts constants and makes them theme-able. LGTM.

Only two comments:
- should we update the Theme here? https://github.com/glideapps/glide-data-grid/blob/main/packages/source/src/stories/use-data-source.stories.tsx
- Is introducing the bubble properties but not making them optional a breaking change?

lukasmasuch:
> Is introducing the bubble properties but not making them optional a breaking change?

I double checked it, but I think we are fine on that aspect since we fill it with a value if it's not provided. The optional or not mainly means if it's guaranteed to be filled with a value when used inside glide-data-grid. For users, all theming props are optional.

Issue #899, opened 2024-02-08
Copying cells containing double quote adds more double quotes


sfc-gh-jipark:
When copying Text cell whose `data` includes a double quote, text copied gets [escaped](https://github.com/glideapps/glide-data-grid/blob/4bdfe38d73b2222f5a4a41426bdbf0da4a26c959/packages/core/src/data-editor/copy-paste.ts#L133) with two more double quotes. 

Note that this happens only when copying into TextBuffer and when copying cell, not text in the textarea in edit mode

What is the rationale for escaping here? Can customize this behavior to respect raw value? Explicitly providing `copyData` doesn't seem to help

https://glideapps.github.io/glide-data-grid/?path=/story/glide-data-grid-dataeditor-demos--copy-support
[screencast-newtab-2024.02.08-12_51_47.webm](https://github.com/glideapps/glide-data-grid/assets/106615738/3c4b9fc9-bf90-4d51-b4a9-846f9aeb7090)

jassmith:
The general rationale here is to make the copy/paste fully compatible (as much as possible) with excel, excel online, google sheets, and apple numbers. The copy buffer contents are both simultaneously well standardized between the different spreadsheets and a total shitshow.

Apple Numbers puts in useless extra newlines, Google Sheets generally is sane, and Excel... oh god... Whatever the fuck those insane people were thinking is beyond me. "Oh hey lets make copy and paste a fucking online service... sounds good!"

That said... yeah I could probably escape less aggressively. I think escaping due to the quotes is technically only needed when the cell contains a newline or a tab character.

bazooka-tango:
How do we turn this off?! No company in my industry copies-and-pastes between Excel and Google Sheet, and any casual user would absolutely intend for copy and paste to actually *copy and paste the contents of the cells without modifying them*. 

This has completely broken every single Google Sheet in my company, we are in dire need of a complete reversion or fix.

PR #1048, opened 2025-06-18
Prevent escaping copy data that was explicitly specified

lukasmasuch:
Don't escape the copy value if the copy data (for copy to clipboard) was explicitly specified by the user via `copyData` cell attribute.

Closes https://github.com/glideapps/glide-data-grid/issues/899


BrianHung (approved):
LGTM; pass raw string back if onCopy is defined. Would add a small test if it's fast but otherwise good.

PR #960, opened 2024-06-01
use pointer events instead of mouse + touch

BrianHung:
Fixes #806 

Not entirely sure why pointer events are more consistent on mobile than touch events.

PR #1050, opened 2025-06-19
Use `pointerdown` instead of `mousedown` for outside click detection

MMihran:
This change replaces the `mousedown` event with `pointerdown` to improve compatibility across input devices. `pointerdown` is part of the Pointer Events API, which provides a unified event model for handling input from various pointing devices — including mouse, touch, and stylus.

Using `pointerdown` ensures better support for touchscreens and other non-mouse inputs.

**Reference:** [MDN – Pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events)


lukasmasuch:
Thanks for the contribution! I think its overlapping with this older PR: https://github.com/glideapps/glide-data-grid/pull/960. 
Planning to merge this one in probably next week, would be awesome if you could take a look at the other PR and add feedback and suggestions.

PR #1053, opened 2025-06-20
Update dependencies - Linaria 6.3.0

danielkita:
Upgrades the CSS-in-JS solution from Linaria 4.5.4 to Linaria 6.3.0 and migrates the build tooling to use @wyw-in-js equivalents, which is the new recommended approach for Linaria v6+.


lukasmasuch (approved):
LGTM 👍

PR #1058, opened 2025-06-22
feat: multi column sort

BrianHung:


lukasmasuch (approved):
LGTM 👍 I added a couple more tests suggested by cursor during reviewing.

PR #1060, opened 2025-06-25
Fix search bar disabled state

lukasmasuch:
The new linaria version seems to require a `&` before css selectors like `:hover` or ":disabled`. This broke the disabled state of the buttons in the search input field.


packages/core/src/internal/data-grid-search/data-grid-search.tsx
```diff
@@ -391,7 +391,7 @@ const DataGridSearch: React.FunctionComponent<DataGridSearchProps> = p => {
 
         return (
             <SearchWrapper
-                className={showSearch ? "" : "out"}
+                className={"gdg-search-bar" + (showSearch ? "" : " out")}
```
lukasmasuch, line 394: Adding a classname here to better target the search bar

PR #1062, opened 2025-06-27
feat: onRowAppended and onColumnAppended

BrianHung:
- adds an `onColumnAppended` function
- allows for `onRowAppended` to be called without trailingRow when new selection would move it beyond row limit

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -218,6 +218,10 @@ export interface DataEditorProps extends Props, Pick<DataGridSearchProps, "image
      * @group Editing
      */
     readonly onRowAppended?: () => Promise<"top" | "bottom" | number | undefined> | void;
+    /** Emitted whenever a column append operation is requested. Append location can be set in callback.
+     * @group Editing
+     */
+    readonly onColumnAppended?: () => Promise<"left" | "right" | number | undefined> | void;
     /** Emitted when a column header should show a context menu. Usually right click.
      * @group Events
      */
@@ -701,6 +705,12 @@ export interface DataEditorRef {
      */
     appendRow: (col: number, openOverlay?: boolean, behavior?: ScrollBehavior) => Promise<void>;
     /**
+     * Programatically appends a column.
+     * @param row The row index to focus in the new column.
+     * @returns A promise which waits for the append to complete.
+     */
+    appendColumn: (row: number, openOverlay?: boolean) => Promise<void>;
+    /**
      * Triggers cells to redraw.
      */
     updateCells: DataGridRef["damage"];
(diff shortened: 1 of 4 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2987,12 +3055,28 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
-                updateSelectedCell(
-                    clamp(gridSelection.current.cell[0] + movX, 0, mangledCols.length - 1),
-                    clamp(gridSelection.current.cell[1] + movY, 0, mangledRows - 1),
-                    isEditingTrailingRow,
-                    false
-                );
+                const isEditingTrailingCol =
+                    gridSelection.current.cell[0] === mangledCols.length - 1 && newValue !== undefined;
+                let updateSelected = true;
+                if (isEditingTrailingRow && movY === 1) {
```
lukasmasuch, line 3061: Maybe its better to do a `onRowAppended !== undefined &&` and keep the old behaviour as is in case `onRowAppended` is not provided?

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2987,12 +3055,28 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                    updateSelected = false;
+                    const col = gridSelection.current.cell[0] + movX;
+                    const customTargetColumn = getCustomNewRowTargetColumn(col);
+                    void appendRow(customTargetColumn ?? col, false);
+                }
+                if (isEditingTrailingCol && movX === 1) {
```
lukasmasuch, line 3067: Same as the other comment with `onColumnAppended !== undefined`

lukasmasuch:
I was able to test the new `onRowAppended` behaviour without a trailing row. But how does it work with `onColumnAppended`? Using `enter` will always go down a row, right?

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -2987,12 +3055,28 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
                 const isEditingTrailingRow =
                     gridSelection.current.cell[1] === mangledRows - 1 && newValue !== undefined;
-                updateSelectedCell(
-                    clamp(gridSelection.current.cell[0] + movX, 0, mangledCols.length - 1),
-                    clamp(gridSelection.current.cell[1] + movY, 0, mangledRows - 1),
-                    isEditingTrailingRow,
-                    false
-                );
+                const isEditingTrailingCol =
+                    gridSelection.current.cell[0] === mangledCols.length - 1 && newValue !== undefined;
```
lukasmasuch, line 3059: nit: does it make sense to rename this to `isEditingLastCol` and `isEditingLastRow` to better differentiate from the `trailingRow` concept?

BrianHung: Updated!

lukasmasuch (approved):
LGTM 👍

PR #1063, opened 2025-06-29
Use pointer events data grid

BrianHung:


citizensas:
> LGMT 👍 I verified that its working in firefox. However, there seems to be an issue with the get mouse args for position story case:
> 
> In firefox:
> 
> image
> In chrome:
> 
> image
> But not 100% sure if its related to this change.

I've noticed the same issue with the getMouseArgsForPosition story on the main branch https://glideapps.github.io/glide-data-grid/?path=/story/glide-data-grid-dataeditor-demos--get-mouse-args-for-position. I don't know what causes it, but it's unrelated to this PR for sure.

lukasmasuch (approved):
LGTM 👍

PR #1067, opened 2025-07-17
feat: compact selection from array

BrianHung:
utility method for creating compact selection from array: useful if you're using toArray and want a fromArray


lukasmasuch (approved):
LGTM 👍

PR #1068, opened 2025-07-18
feat: customizable fill handle config

BrianHung:
Allows fill handle to be customizable. Default is rectangle offset on the corner with outline.

image


BrianHung:
Address copilot comments.

PR #1070, opened 2025-07-18
feat: draw grid lines for selected rows cols

BrianHung:
Explicit borders for column and row selection to match Excel. Default false so users have to opt-in.

CleanShot 2025-07-18 at 15 42 07@2x


BrianHung:
Might update PR to show grid lines in between singleton column selections, as opposed to continuous column ranges.

PR #1073, opened 2025-07-24
feat: expose compact selection items and create static method

BrianHung:
useful if serializing compact selection to json or trying to persist a selection

I'm guessing Jason originally wanted to abstract away how row + column selections are handled internally by glide, but I've found `SelectionRange` to work well when I need a way to persist and handle selections externally.


lukasmasuch (approved):
LGTM 👍 but has a merge conflict to resolve

Issue #1075, opened 2025-07-29
Bug: onClick and onPointerUp both trigger action callbacks for column group actions


hakobyanheghine:
When using column grouping with actions, clicking on a column group action results in the callback being fired twice — once for `onClick` and once for `onPointerUp`.

**To Reproduce**

- Implement getGroupDetails and add actions for column groups.
- Provide a callback for each action (e.g. onClick).
- Click the group action icon.
- Observe the callback is executed twice.

**Root cause**
The issue stems from this code section where both `onClick` and `onPointerUp` handlers invoke groupAction:

In `onPointerUp` handling is [like this](https://github.com/glideapps/glide-data-grid/blob/e7d8fce6d006ca60377259967b93a1ece0b640a7/packages/core/src/internal/data-grid/data-grid.tsx#L1161-L1167):

<img width="822" height="157" alt="Image" src="https://github.com/user-attachments/assets/ea67a5dd-b698-402c-a3c5-3f7a4412b4e1" />

In `onClickImpl` handling is [like this](https://github.com/glideapps/glide-data-grid/blob/e7d8fce6d006ca60377259967b93a1ece0b640a7/packages/core/src/internal/data-grid/data-grid.tsx#L1222-L1226): 

<img width="801" height="122" alt="Image" src="https://github.com/user-attachments/assets/4446d148-7da8-44bb-880c-2a02199a969c" />

So, in both cases the `onClick` is being called

PR #1078, opened 2025-08-05
fix(#1075): groupHeader action click event triggers twice

citizensas:
This PR fixes the #1075 issue. Basically, it removes the `click` event handler for the group header action click and the `pointerUp` event handles the callback function.


BrianHung (approved):
LGTM: group header actions already handled in pointerUp so we dont need it for click.

PR #1080, opened 2025-08-09
bug: getSkipPoint never returns number type value

ImmortalRabbit:
It seems `getSkipPoint` suppose to return either `number` or `undefined` but it's actually never returns `number`


lukasmasuch (approved):
LGTM 👍

PR #1081, opened 2025-08-09
perf: Remove pointless memory allocation for gridSelectionHasItem

ImmortalRabbit:
Remove pointless allocation for gridSelectionHasItem according to FIXME


islittlezhou:
why not mergerd ?

PR #1086, opened 2025-08-14
chore: enable Dependabot for all packages

ImmortalRabbit:
Introducing dependabot.

To enable it, someone with admin access needs to enable dependebot to do security updates.

Security -> Advanced Security - Enable Dependabot

Tested on fork: https://github.com/ImmortalRabbit/glide-data-grid/pulls

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,45 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: "deps"
+      include: "scope"
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "npm"
+    directory: "/packages/core"
+    schedule:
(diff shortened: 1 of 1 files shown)


ImmortalRabbit:
Hi @lukasmasuch 

Do you happen to have admin rights to do so?

ImmortalRabbit:
> > Do you happen to have admin rights to do so?
> 
> Nope :( is this just about security updates, or is this setting also needed to allow normal dependabot updates?

Okay, sorry I think I misunderstood how dependebot works.

So basically yml file should only trigger version updates.

While to enable Security Updates, we should do: `Settings -> Advanced Security - Enable Dependabot`, which we can ignore and just merge yml file only.

As I understand it should create first PRs within first 24 hours after merge, in my fork I just manually enabled security updates and that's why it flooded with PRs.

lukasmasuch (approved):
LGTM 👍 I simplified it a bit to only a single entry since I believe it will discover the other packages because of the workspace setup (will change it if it doesn't work). I also added a cooldown as well.

PR #1088, opened 2025-08-15
Fix homepage links in cells and source packages

eikhr:
Seems like the folder structure has changed, so the links gave 404 errors


lukasmasuch (approved):
LGTM 👍

PR #1102, opened 2025-08-26
Add support for inclusive selection mode

lukasmasuch:
Adds a new `additive` selection mode that allows to mix different selection types without the need of pressing the multi button.


lukasmasuch:
@codex can you add some unit tests?

PR #1172, opened 2026-01-26
fix: add keyboard support for context menu

annahaa874:
## Problem Statement
The data grid previously supported opening context menus only via mouse right-click.
Keyboard users were unable to open context menus using standard shortcuts.

At first glance, handling the contextmenu event appears sufficient. However, this assumption breaks down on macOS, where keyboard shortcuts do not fire a contextmenu event.

## Root Cause
The issue has two contributing factors:

Split event handling architecture
- Keyboard events are handled on the canvas element
- Context menu events are handled on the scroller div
- Keyboard-triggered context menu events would target the canvas, which had no listener

Platform limitation on macOS
- macOS does not fire contextmenu events from keyboard shortcuts

Relying solely on contextmenu listeners is insufficient

## Changes
- Added a Shift + F10 keybinding to explicitly open the context menu
- The handler Invokes the existing onContextMenu callback with event args
- Added unit tests to verify changes

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -3263,7 +3263,6 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     }
                 }
             }
-            console.log({event, details}, "keys.contextMenu:", keys.contextMenu, overlayOpen )
 
             if (details.didMatch) {
                 cancel();
@@ -3363,6 +3362,34 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     col = Number.MAX_SAFE_INTEGER;
                 } else if (isHotkey(keys.goToFirstColumn, event, details)) {
                     col = Number.MIN_SAFE_INTEGER;
+                } else if (isHotkey(keys.contextMenu, event, details) &&
+                  bounds !== undefined &&
+                  event.location !== undefined
+                ) {
+                  const {
+                    location,
+                    ctrlKey,
+                    metaKey,
+                    shiftKey,
+                  } = event;
+
+                  onContextMenu(
+                    {
+                      kind: "cell",
+                      isFillHandle: false,
+                      isTouch: false,
+                      isEdge: false,
+                      button: 0,
+                      scrollEdge: [0, 0],
+                      localEventX: bounds.width / 2,
+                      localEventY: bounds.height / 2,
+                      location,
+                      bounds,
+                      ctrlKey,
+                      metaKey,
+                      shiftKey,
+                      buttons: 0
+                    }, cancel)
                 } else if (rangeSelect === "rect" || rangeSelect === "multi-rect") {
                   if (isHotkey(keys.selectGrowDown, event, details)) {
                     adjustSelection([0, 1]);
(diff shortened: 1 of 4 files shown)


packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3263,6 +3263,7 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     }
                 }
             }
+            console.log({event, details}, "keys.contextMenu:", keys.contextMenu, overlayOpen )
```
citizensas: Please remove this console log.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3363,23 +3364,44 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                        localEventY: event.bounds.height / 2,
+                        location: [col, row],
+                        bounds: event.bounds,
+                        ctrlKey: false,
+                        metaKey: false,
+                        shiftKey: true,
+                        buttons: 0
+                      }, cancel)
+                  }
                 }
```
citizensas, line 3448: I think the `else if` block should be here. You put it in the `rangeSelect === "rect" || rangeSelect === "multi-rect"` if clause.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3363,23 +3364,44 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                        isEdge: false,
+                        button: 0,
+                        scrollEdge: [0, 0],
+                        localEventX: event.bounds.width / 2,
+                        localEventY: event.bounds.height / 2,
+                        location: [col, row],
+                        bounds: event.bounds,
+                        ctrlKey: false,
+                        metaKey: false,
+                        shiftKey: true,
```
citizensas: All of these properties (location, bounds, ctrlKey, metaKey, and shiftKey) exist in the `event` object. I recommend using those values instead.

The `rowMarkerOffset` needs to be considered for the `col` in `location`. But I think that is already handled in the `location` property in the `event` object.

packages/core/src/data-editor/data-editor.tsx
```diff
@@ -3363,23 +3364,44 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
...
+                    onContextMenu(
+                      {
+                        kind: "cell",
+                        isFillHandle: false,
+                        isTouch: false,
+                        isEdge: false,
+                        button: 0,
+                        scrollEdge: [0, 0],
+                        localEventX: event.bounds.width / 2,
+                        localEventY: event.bounds.height / 2,
```
citizensas: Since this is not important for a keyboard event and there is no mouse pointer position, I suggest setting `localEventX` and `localEventY` to `NaN`.

packages/core/src/data-editor/data-editor.tsx
annahaa874: Since rangeSelect can be "none" | "cell" | "rect" | "multi-cell" | "multi-rect", I placed the new condition above it.

packages/core/src/data-editor/data-editor.tsx
annahaa874: Since this is a keyboard event and there is no mouse pointer position, localEventX and localEventY are used to control where the context menu opens. To ensure it opens in the center of the cell, I set them to the cell’s midpoint.

citizensas (approved):
Thank you.

PR #1176, opened 2026-02-17
Add computeMultilineTextLayout for dynamic row height in custom cell renderers

christinewyhsu:
Custom cell renderers currently have no way to know the rendered height of multiline text without duplicating internal logic, which makes https://github.com/streamlit/streamlit/issues/13504 not possible in Streamlit framework, as it depends on this library.

Added two new exports `computeMultilineTextLayout` and `MultilineTextLayout` types to enable the line height computation as a function that can be used by upstream libraries.

Note that the line height calculation is unchanged in `drawMultiLineText` but abstracted into a new function for the export above.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -40,6 +40,7 @@ export {
     roundedPoly,
     roundedRect,
     drawTextCellExternal as drawTextCell,
+    computeMultilineTextLayoutExternal as computeMultilineTextLayout,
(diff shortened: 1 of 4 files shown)


christinewyhsu:
Hi @lukasmasuch, I created this PR to unblock the Streamlit enhancement at https://github.com/streamlit/streamlit/issues/13504. Could you take a look when you get a chance?

PR #1197, opened 2026-08-10
fix: don't block editOnType for cells inside frozen regions

annahaa874:
Pressing a character key to start editing a cell (`editOnType`) silently does nothing if the selected cell is inside a frozen column or row region. Enter-to-edit and click-to-edit both work fine — only the "just start typing" shortcut is affected, because it's the only activation path that consults `visibleRegionRef`, which by design excludes frozen regions from the tracked scrollable area (`vr.x`/`vr.width` end up equal to the frozen-column count, so any column with `index < freezeColumns` fails the `vr.x > col` check).

This exempts cells inside `vr.extras.freezeRegions` from that check, using the same `isInFreezeArea` pattern already established in `getMangledCellContent`'s strict-mode branch elsewhere in this file.

## Test plan
- Added a regression test confirming type-to-edit activates a cell inside a frozen column outside the tracked scroll region (`freezeColumns={2}`, second frozen column).
- Manually verified the same failure/fix behavior holds for a single frozen column (`freezeColumns={1}`) as well, confirming the bug isn't specific to 2+ frozen columns.
- Verified the new test fails without the fix and passes with it.
- Full existing suite passes (155/155 across `data-editor.test.tsx` and `data-grid.test.tsx`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/data-editor/data-editor.tsx
+++ b/packages/core/src/data-editor/data-editor.tsx
@@ -3518,7 +3518,20 @@ const DataEditorImpl: React.ForwardRefRenderFunction<DataEditorRef, DataEditorPr
                     (!showTrailingBlankRow || row !== rows) &&
                     (vr.y > row || row > vr.y + vr.height || vr.x > col || col > vr.x + vr.width)
                 ) {
-                    return;
+                    // Frozen columns/rows are always rendered regardless of scroll position, so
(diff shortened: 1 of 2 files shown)


islittlezhou:
this request will mergeed to man branch ?

